PR #197, opened 2015-12-04
Added items pkgrsc and SMF

tschaefer:
Hi,

I added two new items to the latest release branch 1.5.x:

* [pkgsrc](https://www.pkgsrc.org/)
* [SMF](https://en.wikipedia.org/wiki/Service_Management_Facility)

to support the package and service managent system of [SmartOS](https://smartos.org/).

Greets.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,6 +3,7 @@ branch = True
 source = bundlewrap
 
 [report]
+ignore_errors = True
 exclude_lines =
     pragma: no cover
     def __repr__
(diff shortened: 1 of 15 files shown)


trehn:
Thanks! A couple of things before this can be merged:

* With a contribution of this size, you should add yourself to `AUTHORS`, thereby signing the CAA.
* The stuff you added to `.gitignore` has nothing to do with this project and belongs in a global gitignore, please remove and rebase.
* `pkg_pkgsrc` adds an undocumented and unnecessary dependency on `awk`.
* Alphabetic ordering has been violated in a number of places (`NEEDS_STATIC`, `doc/bundles.rst`)
* PRs should be done against master. The 1.x series will not have any more feature releases, but I can forward port it if you don't want to.

PR #204, opened 2016-02-01
Fix: Don't crash if a user's 'groups' is None

vain:
Not quite sure: Is `None` appropriate here? Or should it be deleted?


trehn:
I think it should be deleted, checking...

trehn:
Dammit, there's a larger problem: Since all attributes for users are optional and unset attributes don't appear in cdicts, a simple user can have a cdict of `{}`. This - by convention - however means that the item should be deleted. Design fail on my part. Thinking...

trehn:
f86304e0feb6c3a7dbd8317cefc41753a73ff884 should fix this

PR #206, opened 2016-02-02
Use ioctl() to detect terminal width and refresh on each update

vain:
As discussed recently.

We assume that `stty size` is too slow to be run on each update. Thus, we directly issue an `ioctl()` call instead of forking/execing.

### DIFF
--- a/src/bundlewrap/utils/ui.py
+++ b/src/bundlewrap/utils/ui.py
@@ -1,12 +1,14 @@
 from codecs import getwriter
 from contextlib import contextmanager
 from datetime import datetime
+import fcntl
 from functools import wraps
 from multiprocessing import Event, Lock, Queue
 import os
 from signal import signal, SIGPIPE, SIG_DFL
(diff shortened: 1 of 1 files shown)


trehn:
Thanks!

I'm still able to mess up the output when repeatedly resizing the terminal, but that already happened before:

http://files.trehn.com/Screen-Shot-2016-02-02-18-49-30-4ofRPc90.png

trehn:
Also I really need to figure out how to fix Travis :/

vain:
Yeah, all sorts of weird things can happen here because we don't draw full frames which contain the entire screen. We just draw parts of the screen and *hope* that the terminal is sane enough to not fsck up the rest of the screen.

What you describe appears to be related to your terminal *rewrapping* its content on resize events. My terminal does that, too, but XTerm for example does not. I can go berserk on resizing an XTerm, bw's status line won't break.

Moral of the story: Draw full frames or live with partially broken output. :-)

PR #210, opened 2016-02-23
Replace U2503 with U2502

vain:
According to Unicode [0], the arcs (U256D and U2570) are "light",
whereas the vertical bar (U2503) is "heavy". This doesn't match. We
should use the light vertical bar (U2502) instead.

[0]: http://www.unicode.org/charts/PDF/U2500.pdf


trehn:
10 points for remembering to update the docs

PR #211, opened 2016-02-24
Comment on what's going on in operations.run()

vain:
I rewrote that function to not use any Event or custom pipes. It should
have worked but it didn't. Well, it mostly worked. Sometimes it didn't.
I then finally discovered that OpenSSH's master process (when using
multiplexing) inherits some pipes and never closes them -- which breaks
everything I had in mind.

Your original idea (using a busy loop and Event) is just the right thing
to do. It's a complex topic, though. Proof reading would be advisable.

### DIFF
--- a/bundlewrap/operations.py
+++ b/bundlewrap/operations.py
@@ -67,14 +67,28 @@ def run(hostname, command, ignore_failure=False, add_host_keys=False, log_functi
     """
     Runs a command on a remote system.
     """
+
+    # LineBuffer objects take care of always printing complete lines
+    # which have been properly terminated by a newline. This is only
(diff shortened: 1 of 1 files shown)


trehn:
👍🐧

trehn:
Thanks. I was kind of hoping that some of this would magically fix Travis, but apparently we can't have nice things.

vain:
For the record, there's a discussion at (portable) OpenSSH's bugzilla about changing that behaviour:

https://bugzilla.mindrot.org/show_bug.cgi?id=1988

I wouldn't expect a change any time soon, though, since this bug is open since 2012.

CC #235.

Issue #162, opened 2015-04-21
tagging system


trehn:
```
svc_systemv = {
    "postgres": {
        'tags': ["downtime"],
    },
}
```

Allows individual items to be tagged. Tags can also be applied at the bundle level to all items in that bundle.

trehn:
Being able to depend on all items with a given tag might also be useful to allow for "virtual resources": some systems need the plain "nginx" pkg_deb while some might need "nginx-extras". But only one of these packages may be installed at the same time. This makes it unsafe for other bundles (who don't want to duplicate the logic that decides which pkg_apt to use) to depend on pkg_apt:nginx. Instead, both pkg_apts could be tagged "provides-nginx" and items could just depend on "tag:provides-nginx".

timbuchwaldt:
<3

PR #215, opened 2016-02-27
Tags

trehn:
PR for #162


trehn:
@vain Do you want to take a look or should I just merge it?

vain:
Just had a quick look. `TagItem` does not use `DummyItem` as super class. Everything else looks fine to me and works as intended/expected. :)

trehn:
thanks

Issue #212, opened 2016-02-25
Document environment variables


vain:
I couldn't find any trace of them in the docs. (Possibly just blind.)

trehn:
Before/when doing this, we should `s/BW_IDENTITY/BWIDENTITY/g` for consistency.

PR #219, opened 2016-03-03
doc: Add documentation about environment variables

vain:
Yes, we should eliminate the underscore in `BW_IDENTITY`. I didn't want
to introduce a breaking change now, though. Having this documentation is
better than no documentation. :-)


vain:
CC #212

trehn:
Since it was never documented anywhere, I'd be fine with just renaming it. If we keep it around, we can still implement it as a fallback only and make `BWIDENTITY` the official name.

Issue #222, opened 2016-03-11
Discard input on stdin before asking a question


vain:
When doing interactive applys with lots of questions, you tend to press enter and press enter and press enter and press enter ... all day long. It's easy to blindly press that button without thinking properly.

Before reading user input on stdin, we should try to discard any input that's currently in the buffer.

I can look into that. Shouldn't be too hard. In any case, Arch Linux's `pacman` does this, so maybe we can have a look at its source code.

PR #224, opened 2016-03-13
Drain any pending input on stdin before asking questions

vain:
PR for #222.

Won't work on Windows or other non-UNIX-like systems, but neither will `term_width()` and maybe a bunch of other stuff.


trehn:
works for me, thanks

PR #235, opened 2016-03-26
Concurrency makeover

trehn:
This went better than expected. I chose a more traditional approach than asyncio, using Python's GIL-impaired pseudo threads.

Running the same `bw test -p 4 -P 4`:
* current master: 151 seconds, ~2 GB RAM peak
* this branch: 80 seconds, 152 MB RAM peak

Even a big `bw verify` that took 559 seconds on master now completes in just 205 seconds.  

There are a few things left to do:
* handling exceptions (currently there is no way to tell which node caused an exception)
* restoring Python 2.7 compatibility using https://pypi.python.org/pypi/futures
* theres a TODO marker about hook registration that needs investigating

### DIFF
--- a/bundlewrap/bundle.py
+++ b/bundlewrap/bundle.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
-from copy import copy
 from os.path import exists, join
 
 from .exceptions import NoSuchBundle, RepositoryError
@@ -34,18 +33,6 @@ class Bundle(object):
         self.bundle_data_dir = join(self.repo.data_dir, self.name)
         self.bundle_file = join(self.bundle_dir, FILENAME_BUNDLE)
         self.metadata_file = join(self.bundle_dir, FILENAME_METADATA)
-
-    def __getstate__(self):
-        """
-        Removes cached items prior to pickling because their classes are
-        loaded dynamically and can't be pickled.
-        """
-        state = copy(self.__dict__)
-        try:
-            del state['_cache']
-        except KeyError:
-            pass
-        return state
 
     @cached_property
     def bundle_attrs(self):
--- a/bundlewrap/cmdline/__init__.py
+++ b/bundlewrap/cmdline/__init__.py
@@ -102,6 +102,7 @@ def main(*args, **kwargs):
         exit(2)
 
     io.debug_mode = pargs.debug
+    io.activate()
 
     if 'BWADDHOSTKEYS' in environ:  # TODO remove in 3.0.0
         environ.setdefault('BW_ADD_HOST_KEYS', environ['BWADDHOSTKEYS'])
--- a/bundlewrap/cmdline/__init__.py
+++ b/bundlewrap/cmdline/__init__.py
@@ -101,7 +101,7 @@ def main(*args, **kwargs):
         parser_bw.print_help()
         exit(2)
 
-    io.activate_as_parent(debug=pargs.debug)
+    io.debug_mode = pargs.debug
 
     if 'BWADDHOSTKEYS' in environ:  # TODO remove in 3.0.0
         environ.setdefault('BW_ADD_HOST_KEYS', environ['BWADDHOSTKEYS'])
--- a/bundlewrap/cmdline/apply.py
+++ b/bundlewrap/cmdline/apply.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals
 from datetime import datetime
 
 from ..concurrency import WorkerPool
+from ..exceptions import WorkerException
 from ..utils.cmdline import get_target_nodes
 from ..utils.text import bold
 from ..utils.text import error_summary, mark_for_translation as _
@@ -41,11 +42,15 @@ def bw_apply(repo, args):
 
             try:
                 result = worker_pool.get_result()
-            except Exception as exc:
+            except WorkerException as exc:
+                msg = "{}: {}".format(
+                    exc.kwargs['task_id'],
+                    exc.wrapped_exception,
+                )
                 if args['debug']:
-                    yield exc.traceback
-                yield str(exc)
-                errors.append(str(exc))
+                    yield exc.wrapped_exception.traceback
+                yield msg
+                errors.append(msg)
             else:
                 node_name = result['task_id']
                 if (
--- a/bundlewrap/cmdline/apply.py
+++ b/bundlewrap/cmdline/apply.py
@@ -4,9 +4,8 @@ from __future__ import unicode_literals
 from datetime import datetime
 
 from ..concurrency import WorkerPool
-from ..exceptions import WorkerException
 from ..utils.cmdline import get_target_nodes
-from ..utils.text import bold, red
+from ..utils.text import bold
 from ..utils.text import error_summary, mark_for_translation as _
 
 
@@ -25,58 +24,49 @@ def bw_apply(repo, args):
     start_time = datetime.now()
 
     with WorkerPool(workers=args['node_workers']) as worker_pool:
-        results = {}
-        while worker_pool.keep_running():
+        while pending_nodes or worker_pool.workers_are_running:
+            while pending_nodes and worker_pool.workers_are_available:
+                node = pending_nodes.pop()
+                worker_pool.start_task(
+                    node.apply,
+                    task_id=node.name,
+                    kwargs={
+                        'autoskip_selector': args['autoskip'],
+                        'force': args['force'],
+                        'interactive': args['interactive'],
+                        'workers': args['item_workers'],
+                        'profiling': args['profiling'],
+                    },
+                )
+
             try:
-                msg = worker_pool.get_event()
-            except WorkerException as e:
-                msg = "{} {}".format(red("!"), e.wrapped_exception)
+                result = worker_pool.get_result()
+            except Exception as exc:
                 if args['debug']:
-                    yield e.traceback
-                if not args['interactive']:
-                    msg = "{}: {}".format(e.task_id, msg)
-                yield msg
-                errors.append(msg)
-                continue
-            if msg['msg'] == 'REQUEST_WORK':
-                if pending_nodes:
-                    node = pending_nodes.pop()
-
-                    worker_pool.start_task(
-                        msg['wid'],
-                        node.apply,
-                        task_id=node.name,
-                        kwargs={
-                            'autoskip_selector': args['autoskip'],
-                            'force': args['force'],
-                            'interactive': args['interactive'],
-                            'workers': args['item_workers'],
-                            'profiling': args['profiling'],
-                        },
-                    )
-                else:
-                    worker_pool.quit(msg['wid'])
-            elif msg['msg'] == 'FINISHED_WORK':
-                node_name = msg['task_id']
-                if msg['return_value'] is not None:  # node skipped because it had no items
-                    results[node_name] = msg['return_value']
-
-                    if args['profiling']:
-                        total_time = 0.0
-                        yield _("  {}").format(bold(node_name))
-                        yield _("  {} BEGIN PROFILING DATA "
-                                "(most expensive items first)").format(bold(node_name))
-                        yield _("  {}    seconds   item").format(bold(node_name))
-                        for time_elapsed, item_id in results[node_name].profiling_info:
-                            yield "  {} {:10.3f}   {}".format(
-                                bold(node_name),
-                                time_elapsed.total_seconds(),
-                                item_id,
-                            )
-                            total_time += time_elapsed.total_seconds()
-                        yield _("  {} {:10.3f}   (total)").format(bold(node_name), total_time)
-                        yield _("  {} END PROFILING DATA").format(bold(node_name))
-                        yield _("  {}").format(bold(node_name))
+                    yield exc.traceback
+                yield str(exc)
+                errors.append(str(exc))
+            else:
+                node_name = result['task_id']
+                if (
+                    result['return_value'] is not None and  # node skipped because it had no items
+                    args['profiling']
+                ):
+                    total_time = 0.0
+                    yield _("  {}").format(bold(node_name))
+                    yield _("  {} BEGIN PROFILING DATA "
+                            "(most expensive items first)").format(bold(node_name))
+                    yield _("  {}    seconds   item").format(bold(node_name))
+                    for time_elapsed, item_id in result['return_value'].profiling_info:
+                        yield "  {} {:10.3f}   {}".format(
+                            bold(node_name),
+                            time_elapsed.total_seconds(),
+                            item_id,
+                        )
+                        total_time += time_elapsed.total_seconds()
+                    yield _("  {} {:10.3f}   (total)").format(bold(node_name), total_time)
+                    yield _("  {} END PROFILING DATA").format(bold(node_name))
+                    yield _("  {}").format(bold(node_name))
 
     error_summary(errors)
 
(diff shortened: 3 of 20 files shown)


trehn:
ready for review

vain:
Starting with `-p 8 -P 8`, I’m getting quite a few errors on a large `bw verify`, stuff like the following:

```
node1:nginx:file:/etc/nginx/sites-available/webapp: not enough values to unpack (expected 4, got 1)
node2:nginx:file:/etc/nginx/conf.d/ssl-ciphers.conf: list index out of range
node3:users:user:jdoe: list.remove(x): x not in list
```

I only see those with all those workers actually being active, i.e. everything’s fine when verifying single nodes or items. I also cannot reproduce the individual errors – things just happen to fail.

I’ll track them down, unless you say, “ah, shit, I know …” ;)

trehn:
one does not simply use threads...

the diverse nature of the exceptions sounds eerily similar to what I hoped I fixed in c57d2dd... (deepcopy uses pickle internally to duplicate stuff)

there was no single consistent error but all kinds of funny stuff like "cannot create instances of sys.version_info"

so unfortunately, I have no clue

vain:
Another whoopsie:

```
$ bw -d verify -p 8 -P 8 foo
...
  File "/home/void/git/bundlewrap/bundlewrap/cmdline/verify.py", line 81, in bw_verify
    result = worker_pool.get_result()
  File "/home/void/git/bundlewrap/bundlewrap/concurrency.py", line 66, in get_result
    raise WorkerException(exception, task_id=task_id, worker_id=worker_id)
bundlewrap.exceptions.WorkerException: list index out of range

During handling of the above exception, another exception occurred:

  File "/home/void/git/bundlewrap/bundlewrap/cmdline/__init__.py", line 65, in wrapper
    return f(*args, **kwargs)
  File "/home/void/git/bundlewrap/bundlewrap/cmdline/__init__.py", line 148, in main
    for line in output:
  File "/home/void/git/bundlewrap/bundlewrap/cmdline/verify.py", line 88, in bw_verify
    yield exc.wrapped_exception.traceback
AttributeError: 'IndexError' object has no attribute 'traceback'
```

The same would happen on `run` and `apply`.

trehn:
meh, looks like I can only get an exception *object* from the thread pool, but in Python 2 those objects don't hold the full traceback

trehn:
with `-p 8 -P 8` I'm also running into my default open files limit of 256, but most of the open files appear to be pipes used to control the threads

vain:
Couple more things:

-   I’m getting a lot of false negatives. For example, bw says “this path is not a directory, but it should be one!” The path in question is perfectly fine. Debugging showed that `file_output` is an empty string in `bundlewrap/utils/remote.py`:`get_path_type()`. Another issue is that sometimes installed packages are recognized as uninstalled.
-   We’re too fast for OpenSSH multiplexing now. I often see `ControlSocket /home/yaddayadda already exists, disabling multiplexing` (after I added some more debugging print statements) on STDERR of the ssh processes. I assume that we are now so fast that we spawn several OpenSSH processes at once – each of them thinking that he’s a master process because the control socket didn’t exist yet. It’s like running `$ for i in {1..8}; do ssh some.host.example.com /bin/true& done` which yields the same error message.
-   Unfortunately, disabling OpenSSH multiplexing does *not* eliminate all those weird errors, but they do appear *a lot* less often.
-   When disabling multiplexing, I’m getting connection timeouts instead (which are *not* related to the index and tuple errors). Opening one ssh connection per `operation.run()` is almost like a DoS attack. It’s too heavy. And, of course, without multiplexing, things are terribly slow. We *need* multiplexing.

This pretty much sucks because it’s a race condition in OpenSSH and we can’t properly fix that. I can only think of an obvious workaround: Don’t spawn all those ssh processes at once. Something as ugly as this might help:

```python
class Node(object):
    def __init__(self):
        self.first_ssh = True
        self.first_ssh_lock = Lock()

    def run(self, ...):
        self.first_ssh_lock.acquire()
        if self.first_ssh:
            self.first_ssh = False
            run_via_ssh('/bin/true')
            sleep(1)
        self.first_ssh_lock.release()
        ...
```

Yeah, I don’t like that either.

trehn:
Why the `sleep(1)`? Shouldn't the control socket be up by the time `run_via_ssh()` returns?

Also, if we do that, we should probably just go ahead and pass `-o ControlMaster...` on our SSH command line. I doubt it'll be a real issue in terms of backwards compatibility. It can still be disabled per node and globally on the command line. This will provide a way better experience for new users who haven't set it up in their SSH config. Also, it enables us to NOT do the `run_via_ssh("/bin/true")` when multiplexing isn't used.

Regarding the DoS situation without multiplexing, we could implement a per-node limit of item workers that caps `-P` passed on the command line. That seems a little over-engineered though.

That just leaves us with the overall weirdness going on. :(

trehn:
Maybe tomorrow I'll have time to create an alternative implementation based on asyncio and then we can see how that goes...

vain:
> Why the sleep(1)? Shouldn't the control socket be up by the time run_via_ssh() returns?

Errr, yes, of course. Brainfart on my end.

> Also, if we do that, we should probably just go ahead and pass -o ControlMaster... on our SSH command line.

:+1:

> Regarding the DoS situation without multiplexing, we could implement a per-node limit of item workers that caps -P passed on the command line.

Hmm, not sure. The server/network decides how many connections are “too much”. I don’t think we can limit `-P` in a meaningful way.

Frohe Ostern.

trehn:
I meant "per-node limit" as in "configurable per-node". Then again its probably better to configure a default number of item workers on a group/node level in the repo and then be able to *override* that with `-P`...

Frohe Ostern.

trehn:
Initial performance numbers on asyncio are encouraging. 102 seconds vs. 341 on master for a `bw test -p -P 4`. This is without actually using any async IO; just balancing tasks on the item level without any parallelism.

Next up would be `bw verify`, but I haven't done that yet. The problem here is that now we're dealing with stuff that really is slow IO. In order to handle this efficiently, I'd have to convert every function call from `main()` down to every single call to `run()` (where we shell out to `ssh`). "Convert" in this case means changing `def foo():` to `async def foo():` and `result = foo()` to `result = await foo()`. That is a lot of layers. The API for custom items would also be affected. Oh and we'd depend on Python 3.5. There is no sensible backwards compatibility to be had.

Pros:
* Speed
* Memory usage
* the basic WorkerPool API from this branch can be kept
 
Cons:
* requires bump to 3.0 just weeks after 2.0 (though most users won't have to touch their repo)
* codebase littered with `await` and `async def`

trehn:
Ugh, nope.jpg.

The restructuring and noise required for asyncio is a deal breaker. `@cached_property` breaks. Now I *did* write an `@async_cached_property`, but that doesn't solve the problem. Because one property must now only ever be called synchronously **or** asynchronously. You can't asynchronously wait for the result and the get the cached version in synchronous calls. `PathInfo` also breaks. And I can't fix that at all without rewriting the entire thing. And that's just getting started in `bw verify`, so there may very well be more problems like this down the road/stack.

So I guess I'm abandoning that effort for now. Just hoping we can unfuck this threading branch.

trehn:
@vain thereifixedit.jpg... 

The problem goes away when we're not using the whole LineBuffer apparatus (which is a good idea anyway since it is only required for `bw run`).

vain:
Can't confirm fixed yet. A `bw verify` is *extremely* slow now. :-) I'll do more debugging tomorrow morning.

trehn:
wtf, now it just hangs

trehn:
adfghsfghjdghj sfdghhjblsdfgp aidfuhg adsfh agiodfghs

vain:
Debugging shows that bw hangs at the first `ssh_process.communicate()`, i.e. the one running `true`.

`strace` shows that the actual ssh process (the one running `true`) has terminated properly. Python is waiting for the forked master process. And here's why.

OpenSSH does triple forking and calls `setsid()` which looks fine. OpenSSH also redirects stdin and stdout to `/dev/null` before forking the master process. They do *not* touch stderr, however, which means that our Python pipe will be inherited by the master process. Python then waits for this pipe to close, which happens as soon as the OpenSSH master process dies.

Please don't ask me why we haven't seen this before.

Here's a simple C program which tries to emulate OpenSSH's forking behaviour:

```C
#include <fcntl.h>
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>

int
main()
{
    pid_t child;
    int null;

    if (fork() == 0)
    {
        null = open("/dev/null", O_RDWR);
        dup2(null, 0);
        dup2(null, 1);
        close(null);
        if (fork() == 0)
        {
            setsid();
            if ((child = fork()) == 0)
            {
                execlp(
                    "/bin/sleep",
                    "sleep",
                    "5",
                    NULL
                );
                perror("execlp");
                exit(1);
            }
            fprintf(stderr, "\"master\" pid: %d\n", child);
        }
    }

    exit(0);
}
```

And here's Python code to call it:
```python
#!/usr/bin/env python

from subprocess import Popen, PIPE

ssh_process = Popen(
    [
        "./bla3",
    ],
    stdin=PIPE,
    stderr=PIPE,
    stdout=PIPE,
)

print("before")
ssh_process.communicate()
print("after")
```

You'll note that this hangs as well. It takes 5 seconds for `after` to appear. Eliminating `stderr=PIPE` fixes the problem.

So much for the analysis. How can we fix this in bw? Could be as simple as this: Don't use the generic `operations::run()` but a special version which does not redirect stderr. The sad thing is, we won't be able to catch output on stderr -- we never will.

trehn:
Wow.

Can't dig into that right now, will try tonight. Spontaneous thought: If we can't capture stderr anyway, can we use `ssh -t`? That would remove the requirement to remove the `requiretty` option (yay words) from sudoers before using bw. But I'm not sure what the side effects would be.

vain:
I think I know why we're sometimes getting empty reads.

Have a look at https://github.com/bundlewrap/bundlewrap/blob/master/bundlewrap/operations.py#L16. The following can happen:

1.  The thread A sleeps on `select()` and eventuall times out.
1.  The scheduler chooses another thread B to run. Our thread A is paused right after having returned from `select()`.
1.  Data becomes available for thread A (but that thread's not running).
1.  `ssh` quits and we reach line 141 in `run()`, thus we set the quit event and wait for the two threads two join.
1.  The scheduler finally chooses our thread A again. It's now running again.
1.  The thread sees `r` being empty and decides it has to go to the `elif` branch.
1.  The thread sees the quit event being set and exits. Data still in the buffer is lost.

I don't have a fix yet. But, yes, this can still happen.

trehn:
Nice catch. Wouldn't the fix simply be trying to read one more time before `return`ing?

vain:
Not sure yet.

Fun fact, please see my comment here: https://github.com/bundlewrap/bundlewrap/blob/master/bundlewrap/operations.py#L130 Oops. Reinventing the wheel makes learning faster, hum?

vain:
Okay, let's go through this one last time.

We're using `stdout=PIPE`, that's fine. stdin is not a problem either.

For stderr, we can't use `stderr=PIPE` because of the `ControlMaster` and pipe inheritance fsckup. In other words, we will *never* be able to tell, "yeah, there's no more data on stderr" because we are not able to get an EOF. We can only make a good guess about it. Even if we do another `select() + read()` after receiving the quit event, I wouldn't be 100% sure that we can catch all data. Granted, ssh has already exited, but are you *sure* that all data written by ssh is already available on our end of the pipe? We can only be *sure* about that if we were to read an EOF. But we won't get an EOF. It's impossible.

Right now, I don't see a way to get this *right*. The best we can do is "do another `select() + read()` after receiving the quit event".

I'll go through OpenSSH's source code, maybe there's a way to turn off stderr for the master process ...

vain:
Nah, can't be turned off in OpenSSH.

[Here](https://anongit.mindrot.org/openssh.git/tree/ssh.c#n1429) is where stdin and stdout are being closed. Right below, there's a call to `daemon(1, 1);` and that second `1` means `noclose = true`, so the descriptors are not being closed in [`daemon()`](https://anongit.mindrot.org/openssh.git/tree/openbsd-compat/daemon.c).

It was worth a try.

trehn:
I would prefer to undo splitting up `run()` and `run_with_log()`, i.e. also read stdout ourselves instead of using `stdout=PIPE`.

vain:
Why? You wouldn't be able to use the same `output_thread_body()` for both the streams. The version for stdout must wait/read until an EOF occurs, the version for stderr must make a best effort. Standard `stdout=PIPE` does just what we want.

trehn:
We have to have a working thread reader for stdout for `bw run` anyway. So why not use it all the time?

trehn:
I was thinking something like this (everything else as it was before 9a40cd):

```python
def output_thread_body(line_buffer, read_fd, quit_event, read_until_eof):
    while True:
        r, w, x = select([read_fd], [], [], 0.1)
        if r:
            chunk = read(read_fd, 1024)
            if chunk:
                line_buffer.write(chunk)
            else:  # EOF
                return
        elif quit_event.is_set() and not read_until_eof:
            return
```

`read_until_eof` is set to `True` for stdout, `False` for stderr

Spoiler: it hangs, not sure yet why

vain:
Probably hangs because $see_my_long_comment_around_line_145? :)

Yeah, I'm fine with doing things that way.

trehn:
I pushed something. It seems to work. Will probably find the next broken thing in about 10 minutes.

trehn:
Still have to update the comments of course.

vain:
I just ran a large `bw verify` and a small `bw apply -i`, both ran fine. :+1:

trehn:
I look forward to the next time we have to mess with this code.

trehn:
Still broken. I'm starting to believe we need to completely change the way we generate metadata.

Lets say we have a **primary** node that we're interested in because we doing a `bw verify` on it. That **primary** node has a metadata processor like this:

```python
def primary(metadata):
    metadata['foos'] = []    
    for secondary_node in repo.nodes_in_group("secondaries"):
        if 'foo' in secondary_node.partial_metadata:
            metadata['foos'].append(secondary_node.partial_metadata['foo'])
    return metadata
```

But our secondary nodes have a slow metadata processor:

```python
def metaproc(metadata):
    time.sleep(100)
    metadata['foo'] = 5
    return metadata
```

Currently when we're generating metadata for a node, we're running and rerunning them, until no metadata processor **on that node** returns changed metadata. However, there is no means by which the secondary node can notify the primary node that it has completed processing its metadata. The primary will contend itself with whatever it received as `secondary_node.partial_metadata`.

Thinking further outside the box, we can imagine chains of dependencies in node metadata. E.g. metadata for node 1 depends on metadata of node 2, which in turn depends on metadata of node 3.

Since bw cannot however predict or inspect these dependencies, we'd have to bruteforce our way through **all** metadata processors on **all** nodes until **none** of them return changed metadata. It's totally irrelevant which nodes are actually *targeted* in a `bw apply` or `bw verify`. Any single node we might be targeting may depend on metadata on *any other* node in the repo.

So we're running into some very serious performance concerns here.

trehn:
Phew, that seems to fix it. I know, I said that before.

The performance impact is not that significant after all. Around 7s for a certain large repo.

vain:
Alright, appears to be working. I also didn't see a noticeable delay. The "processing dependencies" job is much worse.

(btw, I think `float` should be added to `METADATA_TYPES`.)

trehn:
I deliberately left that out because I'm not sure how float equality holds up when comparing metadata processor results. In my mind `float == float` is always a bad idea. Correct me if I'm wrong.

vain:
Blerch. Yes, you're probably right.

trehn:
Jetzt haben wir den Salat.

Issue #241, opened 2016-04-24
Add support for non-Bourne compatible shells


rullmann:
bw will try to set env variables in a Bourne shell compatible way.
Some shells (e.g. fish) don't support this.

Error message:
<pre>
[2016-04-24 18:48:42.774322] bundlewrap: Non-zero return code (127) running 'true' on '192.168.169.30':

Unsupported use of '='. To run 'sudo' with a modified environment, please use 'env LANG=C sudo…'
fish: LANG=C sudo bash -c true
      ^
</pre>

Testing:
<pre>
> env LANG=C sudo bash -c true
> echo $status
0
> LANG=C sudo bash -c true
Unsupported use of '='. To run 'sudo' with a modified environment, please use 'env LANG=C sudo…'
</pre>

It seems like the guys from Ansible had the same issues:
https://docs.ansible.com/ansible/intro_inventory.html#list-of-behavioral-inventory-parameters

trehn:
I think this is more easily fixed by moving `LANG=C` into `-c`:

```diff
diff --git a/bundlewrap/operations.py b/bundlewrap/operations.py
index 12f0b16..1bf3753 100644
--- a/bundlewrap/operations.py
+++ b/bundlewrap/operations.py
@@ -106,7 +106,7 @@ def run(hostname, command, ignore_failure=False, add_host_keys=False, log_functi
             "-o", "PasswordAuthentication=no",
             "-o", "StrictHostKeyChecking=no" if add_host_keys else "StrictHostKeyChecking=yes",
             hostname,
-            "LANG=C sudo bash -c " + quote(command),
+            "sudo bash -c " + quote("LANG=C " + command),
         ],
         stdin=PIPE,
         stderr=stderr_fd_w,
```

This would make it impossible to reliably react to failures in `sudo` or `bash` as requested in #200. But I'm OK with that. I don't think we want to try and be compatible with multiple node shells as Ansible does. That's the whole point of invoking `bash` for everything we do.

@vain any comments?

PR #242, opened 2016-04-24
fix compatibility with fish

trehn:
see #241 for discussion

### DIFF
--- a/bundlewrap/operations.py
+++ b/bundlewrap/operations.py
@@ -106,7 +106,7 @@ def run(hostname, command, ignore_failure=False, add_host_keys=False, log_functi
             "-o", "PasswordAuthentication=no",
             "-o", "StrictHostKeyChecking=no" if add_host_keys else "StrictHostKeyChecking=yes",
             hostname,
-            "LANG=C sudo bash -c " + quote(command),
+            "sudo bash -c " + quote("LANG=C " + command),
         ],
         stdin=PIPE,
         stderr=stderr_fd_w,


vain:
This should work just fine. Haven't seen any unexpected side effects, `LANG=C` is set as before.

This raises another question: Do we actually need Bash? Maybe we should review the code and switch to a simple `sh` (in 3.0). This should help us improve non-GNU compatibility.

(Instead of simply merging this, I'll just do this: "Approved." :-) I think that model works quite well. Code review still happens, but we avoid hasty or unexpected merges.)

(And sorry for the massive delay.)

trehn:
I put `bash` in there to get a more reliable environment, just using whatever `sh` the node provides seemed a bit scary. E.g. what if /bin/sh is linked to /usr/bin/fish or something? Calling `bash` the way we do now reduces our expectations of the remote shell to "can run commands with args separated by spaces and quotes". Apparently setting env vars is already beyond the common denominator of shells.

We do not need bash for the features, I doubt we'll ever need more than env vars, `&&`, `||`, `;`, `|`, `>`, and `<`.

vain:
Invoking `/bin/sh` should reliably get you a "POSIX-compliant" shell (or something very close). If a user really symlinks `/bin/sh` to `/usr/bin/fish`, I fear for his mental health (assuming that fish has no special treatment for being invoked as `sh` -- no idea, I don't use fish). Many "portable" scripts assume `/bin/sh` to be a POSIX-ish shell. If it's not, a *lot* of things will break.

Setting env vars is *way* beyond the common denominator of shells. :) It only works that way in Korn-shell-like shells (afaik). It works differently in FreeBSD's default shell, csh, for example.

Point is, `bash` is not available on all systems, it's not even guaranteed to be available on GNU. But then again, `sudo` is not available on FreeBSD or NetBSD per default either. Maybe we can't avoid some "abstraction layer" anyway ...

Whatever. Sorry for going off-topic.

Issue #137, opened 2014-09-17
Enabling/disabling of service items


vain:
In #60, I stated some reasons why `svc_systemd` does not support `enable` or `disable`. I'm not so sure about that anymore.

Given BundleWrap's mindset and its current feature set (canned actions!), I think that we should provide a way to simply enable or disable a service. If the user wants fancy stuff, he can still create symlinks by himself.

(This is less relevant for Debian based systems because most daemons are activated by default here. On Arch, though, you'll end up with a lot of manually defined symlinks.)

trehn:
Closing this, open new issues for individual items if you want.

PR #244, opened 2016-05-09
Enable/disable services

vain:
PR for #137. Only systemd for now, so consider this WIP.

Note: This PR is not backwards compatible because the default value for `enabled` is `True`. Hence, when using this with existing repos, all services will automatically be enabled which is not necessarily what the user wants.

### DIFF
--- a/bundlewrap/items/svc_systemd.py
+++ b/bundlewrap/items/svc_systemd.py
@@ -24,27 +24,60 @@ def svc_stop(node, svcname):
     return node.run("systemctl stop -- {}".format(quote(svcname)))
 
 
+def svc_enable(node, svcname):
+    return node.run("systemctl enable -- {}".format(quote(svcname)))
+
+
+def svc_enabled(node, svcname):
+    result = node.run(
+        "systemctl is-enabled -- {}".format(quote(svcname)),
+        may_fail=True,
+    )
+    return result.return_code == 0
+
+
+def svc_disable(node, svcname):
+    return node.run("systemctl disable -- {}".format(quote(svcname)))
+
+
 class SvcSystemd(Item):
     """
     A service managed by systemd.
     """
     BUNDLE_ATTRIBUTE_NAME = "svc_systemd"
     ITEM_ATTRIBUTES = {
+        'enabled': None,
         'running': True,
     }
     ITEM_TYPE_NAME = "svc_systemd"
 
     def __repr__(self):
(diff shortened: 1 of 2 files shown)


vain:
This is a little messy for Upstart:

http://upstart.ubuntu.com/cookbook/#override-files

"Override files" can be used to disable a service. Fine. But they can be used for other purposes as well. Maybe an admin does all kinds of fancy stuff here. I don't feel comfortable with bw touching them.

I see two options:

1.  Don't implement enable/disable for `svc_upstart`. Most services are "enabled" by default anyway, because as soon as an upstart file is dropped in `/etc/init`, it will be used on next boot (if it contains a `start on` line which 99% of them do). If a user wants to disable a service, he can drop an override file manually (or via bw).
1.  We could disable a service by renaming its service file so it doesn't end in `.conf`. I consider this to be not-so-pretty, but hey, it's what the official Upstart documentation suggests.

To be honest, I'd go with option 1. I haven't ever felt the need to enable/disable an Upstart service. On Ubuntu, everything is automatically enabled as soon as you install it ...

Plus, Upstart is dying.

trehn:
Agreed, forget Upstart.

For systemd, the default should be `None` as in "don't care" for backwards compatibility. You'll have to write a custom `cdict()`:

```python
def cdict(self):
    cdict = self.attributes.copy()
    if cdict['enable'] is None:
        del cdict['enable']
    return cdict
```

vain:
Of course, SystemV init is a total mess.

We don't know in which runlevels to start a service. a) An initscript doesn't contain that information, b) different distributions can have different definitions for their runlevels (https://en.wikipedia.org/wiki/Runlevel).

Something like `update-rc.d` is a Debian-only thing.

If anything, we could provide a nicer interface to create symlinks in the `/etc/rc*` directories. But the user would have to tell us which directories to use -- and how. "Create `K50${service}` in `/etc/rc6.d`." Something like that.

Do we want to support this? Could be worth checking how/if other config management systems handle this.

vain:
Regarding systemd's default value for `enabled`: I think the *right* thing to do is use `True` as a default. That's what you want most of the time.

How about using `None` as a "transitional" default value and switching to `True` in bw 3.0?

trehn:
https://github.com/ansible/ansible-modules-core/blob/devel/system/service.py#L686

Scrolling through that file makes me sad.

Issue #191, opened 2015-10-13
`bw lock` to manually lock nodes


trehn:
`-u` to unlock

`-m` for optional message

vain:
How about a "special" kind of lock which can not be overriden by *other* users (via bw -- of course, the user can always manually remove /tmp/bundlewrap.lock)? I myself want to be able to work on the node, but I want to prevent others from interfering.

trehn:
Hm, but how do we tell different users apart? The local username currently in the lock file is fragile, the remote user on the node might be shared.

trehn:
This probably has to be entirely separate from the existing locking system (which also prevents multiple applies from the same user).

vain:
Why do you consider the local username fragile?

I can only think of one scenario: A user works on multiple workstations, each with a different username. Frankly, I don’t think that’s very common.

trehn:
I was more concerned about two persons using the same username on their respective machines ("admin", "ubuntu", ...).

But I guess #194 is sufficient to address this.

vain:
Arrrr. People who do this simply deserve to override/ignore each other’s lock files.

trehn:
Locks should also have a default timeout of 8h.

Keep in mind that we can't break backwards compatibility with existing lock files (since people using the same repo might be on slightly different versions of bw). Therefore, the current lock file format has to stay. I'm undecided whether we should add a default timeout for these existing locks.

Now the problem is that adding a second lock file to the same directory will break all prior versions (they find a lock dir without the file they're looking for). So we have to add a second lock directory. This new kind of lock is of course unknown to and ignored by prior versions of bw.

I propose these new lock file contents:

```json
[
{
    "date_acquired": 1234567890,
    "date_timeout": 9876543210,
    "locked_by": "jdoe@example.com",
    "exempt_from_lock": ["jdoe@example.com"],
    "block": ["apply", "lock"]
}
]
```

Note that it's a list, so we can add additional locks that overlap on a node. This is important so more than one person can place a lock on a node. Why is that a thing? Because with `block` we can specify which actions we want to block: "apply" obviously blocks `bw apply`, while "run" blocks `bw run` and "lock" prevents anybody from placing additional locks on the node. This makes it possible for independent users to disable applying configuration and exempt the nodes from large scale `bw run`s.

CLI:

```
$ bw lock --unlock NODE|GROUP  # removes all locks from node owner by current user
[...]

PR #246, opened 2016-05-16
Soft locks

trehn:
PR for #191 

The existing lock is now called a "hard lock" because it continues to prevent concurrent apply operations on the node, regardless of which user it is. These new locks are "soft locks" because they are ignored by the users who created them.

As of now, the management UI for the soft locks is in place. TODO:

* [x] remove expired locks in `softlock_list()`
* [x] check `softlock_list()` at the beginning of `bw run` and `bw apply`
* [ ] think about a good default expiry timeframe, maybe 12h is better than 8h
* [x] docs
* [ ] tests

### DIFF
--- a/bundlewrap/cmdline/lock.py
+++ b/bundlewrap/cmdline/lock.py
@@ -81,8 +81,10 @@ def bw_lock_show(repo, args):
         return
 
     for lock in locks:
-        lock['formatted_date'] = datetime.fromtimestamp(lock['date']).strftime("%c")
-        lock['formatted_expiry'] = datetime.fromtimestamp(lock['expiry']).strftime("%c")
+        lock['formatted_date'] = \
+            datetime.fromtimestamp(lock['date']).strftime("%Y-%m-%d %H:%M:%S")
+        lock['formatted_expiry'] = \
+            datetime.fromtimestamp(lock['expiry']).strftime("%Y-%m-%d %H:%M:%S")
         lock['formatted_ops'] = ", ".join(sorted(lock['ops']))
 
     headers = (
--- a/bundlewrap/cmdline/lock.py
+++ b/bundlewrap/cmdline/lock.py
@@ -1,10 +1,12 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
+from datetime import datetime
+
 from ..concurrency import WorkerPool
-from ..lock import softlock_add
+from ..lock import softlock_add, softlock_list
 from ..utils.cmdline import get_target_nodes
-from ..utils.text import bold, error_summary, green, mark_for_translation as _
+from ..utils.text import blue, bold, cyan, error_summary, green, mark_for_translation as _
 from ..utils.ui import io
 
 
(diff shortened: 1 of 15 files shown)


trehn:
One remaining problem with a busy pull request workflow is that while I can now lock certain nodes and make them accessible to *me*, that doesn't lock these nodes to other *branches* I might be working on. But I believe the only way around that would be deep git integration (no) or manually "tagging" each `bw apply` with a branch name (no).

vain:
Currently reviewing.

> One remaining problem with a busy pull request workflow is that while I can now lock certain nodes and make them accessible to me, that doesn't lock these nodes to other branches I might be working on.

Let's encourage the users to cheat and brutally abuse `$BW_IDENTITY` (kidding, kind of).

vain:
Buggy McBugface:

```
# l bundlewrap.softlock.d/
total 12,288
drwxr-xr-x 2 root     root      4,096 May 16 18:43 ./
drwxrwxrwt 4 root     root      4,096 May 16 18:43 ../
-rw------- 1 lulufant hellyeah    135 May 16 18:43 KAQE
```

But:

```
        cat = node.run("cat {}".format(SOFT_LOCK_FILE.format(id="*")), may_fail=True)
        if cat.return_code != 0:
            return []
```

Meaning, when another user puts a lock on the system, my `cat` will yield a `Permission denied` with exit code 1, so my `bw` ignores the other user's lock.

Easy to fix, just upload locks with `mode=0644`.

vain:
:+1: for everything else. This is going to be a huge improvement. :)

vain:
>> One remaining problem with a busy pull request workflow is that while I can now lock certain nodes and make them accessible to me, that doesn't lock these nodes to other branches I might be working on.

> Let's encourage the users to cheat and brutally abuse `$BW_IDENTITY` (kidding, kind of).

On a more serious note: https://github.com/vain/bin-pub/commit/e403b0fdb8d4d03b879139988407991984159a93

I don't think there's a way to avoid "local tooling" like this, unless we want to support each and every VCS (including NO VCS) in bw.

trehn:
Random idea: Maybe instead of the current choice between "run", "apply", or both we can just add item selectors to locks. One could still lock down a node completely by using `*`, but it would be possible to prevent changes to specific "small" items like zonefiles or just a certain bundle (with `bundle:foo`) while still allowing the rest of the node to be managed. The locked items would just be skipped.

I'm not sure there is a need to prevent `bw run`.

trehn:
@vain care for a final review?

vain:
Works like a charm.

I've also added support for `$BW_SOFTLOCK_EXPIRY`. Want me to push it? Or is that a bad idea? People might set this to "three weeks" so they can be lazy. But it would allow for a better adaption to the team's workflow ...

trehn:
By all means, push it. Was just too tired to add that on top. I have another almost complete commit that will show the default defaults or the defaults from env vars in `--help` for all commands. Add your stuff, then I can complete that.

PR #247, opened 2016-05-19
[WIP] OpenBSD compatibility

timbuchwaldt:
This pull request introduces support for OpenBSD.

Todo:

- [x] Documentation
- [x] CLA
- [x] Add pkg_openbsd
- [x] Add svc_rcd

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -3,4 +3,4 @@
 
 Torsten Rehn <torsten@rehn.email>
 Peter Hofmann <scm@uninformativ.de>
-Tim Buchwadlt <tim@buchwaldt.ws>
+Tim Buchwaldt <tim@buchwaldt.ws>
--- a/AUTHORS
+++ b/AUTHORS
@@ -3,3 +3,4 @@
 
 Torsten Rehn <torsten@rehn.email>
 Peter Hofmann <scm@uninformativ.de>
+Tim Buchwadlt <tim@buchwaldt.ws>
--- a/bundlewrap/items/directories.py
+++ b/bundlewrap/items/directories.py
@@ -73,7 +73,11 @@ class Directory(Item):
                 getattr(self, "_fix_" + fix_type)(status)
 
     def _fix_mode(self, status):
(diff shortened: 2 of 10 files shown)


timbuchwaldt:
I would suggest the following to refactor the inline if-switches on top of all items that need it:
```python
def os_dependent(os):
    if os == "openbsd":
        return {
           "chown_command": "chown {}{} {}"
        }
    elif os == "otheros":
        return {
           "chown_command": "chown {}{} {}"
        }
    else:
        return {
           "chown_command": "chown {}{} -- {}"
        }
```

In the logic itself the if/else branches would go away in favor of `os_dependent(self.node.os)["chown_command"]`

trehn:
Is there a good reason for changing the non-OpenBSD `_DEFAULT_SALT`?

The docs for `use_shadow` should reflect that it does nothing when `node.os` is `openbsd`.

trehn:
The new items must also be added to `docs/content/repo/bundles.md`.

timbuchwaldt:
RE changing the default salt: Nope, was trying out and forgot. Will update the docs.

vain:
Gave it a try on an OpenBSD 5.9 VM, looks good to me. :)

Two details:

-   `svc_rcd.py`, function `svc_enabled()`: Wouldn't it be better to use `"rcctl ls on | grep ^{}$"` to avoid partial matches?
-   `svc_rcd.md`: The docs still contain systemd examples, i.e. `fcron.service` and `sgopherd.socket`. That might be a little confusing.

(Maybe we can find a better term than `svc_rcd`. `/etc/rc.d` is available on many systems, including GNU. They all work slightly different, though.)

timbuchwaldt:
@vain Cool! :)

Will change both. The slight differences concern me, too. Maybe svc_openbsd would be more clear, as it states on which OS it works. Or adding compatibility to e.g. Gentoo - but I have no motivation to do that TBH.

trehn:
Thanks! :lovetim:

PR #250, opened 2016-06-03
Use `grep -a` for /etc/passwd on GNU

vain:
The full name ("gecos") of a user might contain non-ASCII characters. On
systems with `LANG=C`, this can lead to grep saying:

    Binary file matches

(Instead of the actual match.)


vain:
I don't know how/if this applies to OpenBSD as well. Will have to check.

trehn:
same on OpenBSD

Issue #252, opened 2016-06-10
allow bw commands recursively in repo directory


bkendinibilir:
It would be helpful, if the bw command can also be used in subdirectories of the bw repo directory. Vagrant allows this, too.

PR #254, opened 2016-06-13
invoke bw from repo subdirs

trehn:
PR for #252

### DIFF
--- a/bundlewrap/cmdline/__init__.py
+++ b/bundlewrap/cmdline/__init__.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 
 from functools import wraps
 from os import environ, getcwd
+from os.path import dirname
 from sys import argv, exit, stderr, stdout
 from traceback import print_exc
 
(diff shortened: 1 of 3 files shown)


vain:
Works with a simple test repo, but some repos out there might rely on CWD being the root of the repo. In a way, this breaks backwards compatibility, doesn't it?

One might think of doing a `chdir(path)` after the `path = dirname(path)`. This works when invoking bw from the command line.

I just realized the whole topic gets pretty confusing once you start using bw as a library. Right now, you can't just do a `r = Repository('/my/path')` -- you manually have to issue a `chdir()` first.

To be honest, we should either consider repos that depend on a specific CWD to be broken, or we should make `Repository` robust enough to handle these cases. I vote for the first option.

vain:
> If you're talking about BundleWrap itself, that's a bug.

Nah, I was referring to repos that depend on CWD.

Alright then. Depending on CWD is bad. :)

Issue #75, opened 2014-02-22
abort gracefully on CTRL+C


trehn:
CTRL+C again for hard abort

vain:
I played with the idea we discussed today.

This prevents ssh from receiving `SIGINT` from the terminal and also catches it in our parent process:

```python
#!/usr/bin/env python

from os import setpgrp
from subprocess import PIPE, Popen
from signal import SIGINT, signal

def new_process_group():
    setpgrp()

def print_sigint(*args, **kwargs):
    print("SIGINT IN PYTHON")

signal(SIGINT, print_sigint)

p = Popen(
    ['ssh', 'my.node.com', 'sleep 10m'],
    preexec_fn=new_process_group,
    stdin=PIPE,
    stdout=PIPE,
    stderr=PIPE,
)
print(p.pid)
p.communicate()
print("Process finished")
```

Now, normally, our thread will block in `p.communicate()` and only return from that function if ssh exits. That's fine and that's what you'd expect from a "soft shutdown".

For a "hard shutdown", we'd have to manually send all of our ssh processes a signal. I simulated this by manually doing a `kill $sshpid` on the shell. Works fine, the code above returns from `p.communicate()` and the ssh process is gone. This means we'd have to collect all ssh PIDs in our main thread in order to send them a signal in case of a hard shutdown.

There's no easy way out here, like "kill all processes in our process group". That's exactly what `setpgrp()` prevents: It creates a new process group just for ssh, so it won't receive the `SIGINT` that's implicitly sent to the foreground process group. Also, there is no way to send a signal to all children.

Note that ignoring `SIGINT` in before `exec()`ing ssh is useless, too: ssh resets the signal handler. Instead, we must take care that it never receives the signal in the first place.

And, another idea: We could just kill all of bw's threads, wouldn't this kill ssh, too? No, it doesn't.

(Careful, that little snippet above can easily fool you. It's just that "ssh multiplexing master doesn't close stderr" thingy again. If your ssh appears to never return, that's why.)

vain:
Aww, too bad, multiplexing breaks everything. Even when there's already a master running.

I'm ending up with an ssh zombie. Python doesn't care about the `SIGCHLD` it receives.

Will have to investigate ...

PR #258, opened 2016-06-29
[WIP] Graceful abort on ^C

vain:
PR for #75.

Open for feedback, even though it's still a little WIP-ish.


trehn:
I am strongly in favor of more tightly integrating the SIGINT handler with the existing IO system. I have already replaced the direct stderr write with `io.stderr` on the branch, breaking `ask()`.

I think we should also be able to clean up our status line during a hard shutdown just before `_exit()`ing.

Why would using `QUIT_EVENT` in the handler be a problem?

Just thinking: Is SIGINT really the appropriate signal to pass on to SSH? Are we not more determined at this point so we could send SIGTERM? Will have to check if SSH passes that on to the remote process on the node, too.

It would be nice to not have "^C" mess up our output, but turning off echo for that seems excessive.

trehn:
When CTRL+C works during ask(), we should probably also get rid of the "q" answer, keeping with the Python spirit of there being only one way of doing something.

trehn:
That was easier than I expected.

But for some reason, `git_deploy` complains during a *soft* shutdown:

```
i SIGINT  canceling pending tasks... (hit CTRL+C again for immediate dirty exit)
failed command: git clone --bare --depth 1 --no-single-branch https://github.com/MalcolmRobb/dump1090 .
stdout:
b''
stderr:
b"Cloning into bare repository '.'...\n"
  File "/Users/trehn/Projects/private/software/bundlewrap/bundlewrap/concurrency.py", line 133, in run
    result = self._get_result()
  File "/Users/trehn/Projects/private/software/bundlewrap/bundlewrap/concurrency.py", line 73, in _get_result
    raise exception
  File "/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/concurrent/futures/thread.py", line 55, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/Users/trehn/Projects/private/software/bundlewrap/bundlewrap/node.py", line 474, in apply
    profiling=profiling,
  File "/Users/trehn/Projects/private/software/bundlewrap/bundlewrap/node.py", line 202, in apply_items
    worker_pool.run()
  File "/Users/trehn/Projects/private/software/bundlewrap/bundlewrap/concurrency.py", line 142, in run
    raise exc
  File "/Users/trehn/Projects/private/software/bundlewrap/bundlewrap/concurrency.py", line 133, in run
    result = self._get_result()
  File "/Users/trehn/Projects/private/software/bundlewrap/bundlewrap/concurrency.py", line 73, in _get_result
    raise exception
  File "/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/concurrent/futures/thread.py", line 55, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/Users/trehn/Projects/private/software/bundlewrap/bundlewrap/items/__init__.py", line 379, in apply
    status_before = self.cached_status
  File "/Users/trehn/Projects/private/software/bundlewrap/bundlewrap/utils/__init__.py", line 43, in cache_wrapper
    return_value = prop(self)
  File "/Users/trehn/Projects/private/software/bundlewrap/bundlewrap/items/__init__.py", line 205, in cached_status
    return self.get_status()
  File "/Users/trehn/Projects/private/software/bundlewrap/bundlewrap/items/__init__.py", line 563, in get_status
    return ItemStatus(self.cached_cdict, self.cached_sdict)
  File "/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/contextlib.py", line 77, in __exit__
    self.gen.throw(type, value, traceback)
  File "/Users/trehn/Projects/private/software/bundlewrap/bundlewrap/utils/ui.py", line 172, in job
    yield
  File "/Users/trehn/Projects/private/software/bundlewrap/bundlewrap/items/__init__.py", line 563, in get_status
    return ItemStatus(self.cached_cdict, self.cached_sdict)
  File "/Users/trehn/Projects/private/software/bundlewrap/bundlewrap/utils/__init__.py", line 43, in cache_wrapper
    return_value = prop(self)
  File "/Users/trehn/Projects/private/software/bundlewrap/bundlewrap/items/__init__.py", line 175, in cached_cdict
    cdict = self.cdict()
  File "<string>", line 156, in cdict
  File "/Users/trehn/Projects/private/software/bundlewrap/bundlewrap/utils/__init__.py", line 43, in cache_wrapper
    return_value = prop(self)
  File "<string>", line 142, in _expanded_rev
  File "/Users/trehn/Projects/private/software/bundlewrap/bundlewrap/utils/__init__.py", line 43, in cache_wrapper
    return_value = prop(self)
  File "<string>", line 148, in _repo_dir
  File "<string>", line 46, in clone_to_dir
  File "<string>", line 111, in git_command

RuntimeError('`git clone` failed in /var/folders/5x/b7p2m3xs0p9gjn73y4nd26dr0000gn/T/tmp01uwkzqf',)
pi-1: `git clone` failed in /var/folders/5x/b7p2m3xs0p9gjn73y4nd26dr0000gn/T/tmp01uwkzqf
```

vain:
> I have already replaced the direct stderr write with io.stderr on the branch, breaking `ask()`.

Does the main thread ever use `io.stderr()`? If yes, then we can deadlock:

-   Main thread calls `io.stderr("hi")`.
-   `io.stderr` acquires the IO lock.
-   User hits ^C, signal handler is called (on the main thread), while that other `io.stderr` still holds the lock.
-   We call `io.stderr` from the signal handler, this tries to acquire the lock which will never be freed again.

If it's guaranteed that only other threads will call `io.stderr()`, then we're okay. Our signal handler will still block when acquiring the lock, but the other thread continues to run and will eventually free the lock.

That's also why I'm not happy with using `QUIT_EVENT` in the signal handler. Does `threading.Event` internally use locks? Can this cause a deadlock? I haven't checked yet, sorry.

> Just thinking: Is SIGINT really the appropriate signal to pass on to SSH? Are we not more determined at this point so we could send SIGTERM?

Yup, SIGTERM makes more sense.

trehn:
Unfortunately I had to move all the signal handling back in with the IO stuff for import reasons :( But the deadlock should be dealt with now. 

I looked at the source of Event (it's pretty simple actually) and it's not a problem. SIGTERM is also used now.

Looks all good to me except for the `git_deploy` issue, which I still don't understand.

vain:
Yes, looks good!

> Looks all good to me except for the `git_deploy` issue, which I still don't understand.

`git` still receives the original SIGINT because it doesn't live in its own process group. I opened a (very simple) PR in the plugin repo to fix that.

trehn:
Of course, thanks.

Issue #186, opened 2015-08-31
better support for "empty" nodes


trehn:
it's often useful to have nodes without configuration, e.g. for monitoring of hardware alongside monitoring for standard bw nodes

not assigning any bundles still has bw trying to connect during `bw run` and `bw apply`

so there must be an option to declare a node (including metadata), but explicitly forbid connecting to it

ideas:

```python
nodes = {
    "mynode": { 'connect': False },
}
``` 

```python
nodes = {
    "mynode": { 'virtual': True },
}
``` 
I'm leaning towards the latter. Are there any other use cases for "virtual" nodes like this?

trehn:
note that people still may want to use `bw run` for those nodes, just not `bw apply`

trehn:
virtual nodes should just live in metadata

PR #259, opened 2016-06-30
node.dummy

trehn:
CC #186

### DIFF
--- a/bundlewrap/cmdline/nodes.py
+++ b/bundlewrap/cmdline/nodes.py
@@ -2,6 +2,7 @@
 from __future__ import unicode_literals
 
 from ..utils import names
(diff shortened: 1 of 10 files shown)


trehn:
I think there is more to do (not necessarily in this PR though) wrt node attributes. `bw nodes --attrs` works well enough with `grep` and `cut`, but isn't very human-friendly. And it will become even less so as we add more attributes (e.g. `node.os_flavor` or `node.os_version`).

Also, it might become desirable to build groups based on attributes, so users can easily apply on all Ubuntu nodes still running 12.04 for example. No idea how to do that yet.

vain:
:+1:

As for the output: Hooray for multiline output!

PR #260, opened 2016-07-02
add BW_SSH_ARGS

trehn:


vain:
I hope my life stays simple enough, so I never need to use this with `-F`.

Works fine, though. :v:

PR #262, opened 2016-07-04
drop requirement for bash on nodes

trehn:
In the end, it probably does more harm than good, especially wrt BSD support.

related #241, #242


vain:
As we're not doing any fancy Bash-istic stuff, this should be just fine. (We probably should have done this a long time ago.)

Issue #202, opened 2016-01-01
Add support for RedHat DNF


rullmann:
`dnf` will replace `yum` in the future.
Currently using the `yum`-action doesn't result in an error as yum "redirects" to dnf.

`dnf` should be an item in bw in the future.

https://fedoraproject.org/wiki/Features/DNF

sq3:
I started working on this a few weeks ago, but it's currently untested

https://github.com/sq3/bundlewrap/commit/717406cd864e00b7481c7442b120d8c739d4abc1
https://github.com/sq3/bundlewrap/commit/85fc99ca5abc3d059ddd6f87dd4e2f0c39244ddd

trehn:
To me the linked wiki page reads like DNF is slated to be a drop-in replacement for yum and will adopt the yum name once it matures. The `dnf` executable appears to only be a temporary thing until the old yum goes away. If that is the case, I don't see any reason why bw would want to support this. Am I reading that wrong?

PR #281, opened 2016-10-23
add dnf support

rullmann:
clone pkg_yum to support dnf, the next generation yum

### DIFF
--- a/bundlewrap/items/pkg_dnf.py
+++ b/bundlewrap/items/pkg_dnf.py
@@ -31,7 +31,7 @@ class DnfPkg(Item):
     """
     A package installed by dnf.
     """
-    BLOCK_CONCURRENT = ["pkg_dnf"]
+    BLOCK_CONCURRENT = ["pkg_dnf", "pkg_yum"]
     BUNDLE_ATTRIBUTE_NAME = "pkg_dnf"
     ITEM_ATTRIBUTES = {
         'installed': True,
--- a/bundlewrap/items/pkg_dnf.py
+++ b/bundlewrap/items/pkg_dnf.py
@@ -1,0 +1,66 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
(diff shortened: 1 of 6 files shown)


rullmann:
Lokal mit python3 gebaut und getestet.

bundlewrap/items/pkg_dnf.py
```diff
@@ -0,0 +1,66 @@
...
+
+def pkg_remove(node, pkgname):
+    return node.run("dnf -d0 -e0 -y remove {}".format(quote(pkgname)))
+
+
+class DnfPkg(Item):
+    """
+    A package installed by dnf.
+    """
+    BLOCK_CONCURRENT = ["pkg_dnf"]
```
trehn: Should this not block `pkg_yum`?

trehn: Also, what about collisions between `pkg_yum` and `pkg_dnf`? Is it OK to have `pkg_yum:foo` and `pkg_dnf:foo` on the same node?

trehn: Feel free to add yourself to AUTHORS if you want.

trehn:
Feel free to add yourself to AUTHORS if you want.

trehn:
CC #202

bundlewrap/items/pkg_dnf.py
rullmann: Need to learn more about blocking. I'm worried that in a mixed mode (old and new bundles) this will break something?
But yes, it's oke to have both on one node, as it won't break anything, only adds an unnecessary check.

rullmann:
Regarding #202: `/usr/bin/yum` on Fedora is only a bash script redirecting to dnf. yum as command is deprecated.
I haven't seen any discussions about replacing the yum command itself with the dnf binary.

trehn:
Thx!

Issue #284, opened 2016-11-10
allow more flexible composition of groups


trehn:
determine group membership by any of these
* a certain value for an attribute (`os`, `os_version`, `dummy`)
* intersection between two other groups
* member of one other group, but not of a third group

vain:
Opening pandora's box: Group membership processors?

trehn:
THE BOX HAS BEEN OPENED

One problem is that node attributes are often set by groups in the first place and are therefore hard (though not impossible) to use for group assignment. Groups assigned through attributes then could set new attributes... In the end, anything that selects group membership based on attributes such as `os` or `dummy` would mean additional complexity about the size of metadata processors. Since we're adding regular subgroups and loop detection on top, such processing would probably be even more complicated than metadata processors.

trehn:
That being said, right now the only way to get an Ubuntu group is to do this:

```python
groups = {
    "ubuntu": {
        'os': 'ubuntu',
        'members': [
            "node1",
            "node2",
        ],
    }, 
}
```

Which is obviously a huge pain. You always want to set the OS on the node itself.

So there is quite something to be gained from attribute-based groups.

trehn:
What we could do is this:

```
def select_ubuntu(nodes):
    for node in nodes:
        if node.os == 'ubuntu':
            yield node

groups = {
    "ubuntu": {
        'select_nodes': [
            select_ubuntu,
        ],
    }, 
}
```

The important question here is: do we run the functions in `select_nodes` repeatedly like metadata processors every time an attribute changes on a node?

trehn:
Maybe a simple boolean test would be more elegant:

```python
groups = {
    "ubuntu": {
        'select_nodes': [
            lambda node: node.os == 'ubuntu',
        ],
    }, 
}
```

PR #285, opened 2016-11-11
add group.members_add and .members_remove

trehn:
#284


vain:
Played with this for a while. Haven't found any nasty corner cases so far. :+1:

(Code turned out to be simpler than I expected. :-))

trehn:
thanks for checking it out

Issue #274, opened 2016-08-18
always write debug log file


trehn:
(if env var with base path is set)

user has to take care of rotation and cleanup

avoids having to re-run with `--debug` after something went wrong

PR #294, opened 2016-11-24
add `BW_DEBUG_LOG_DIR`

trehn:
#274

### DIFF
--- a/bundlewrap/utils/ui.py
+++ b/bundlewrap/utils/ui.py
@@ -3,7 +3,8 @@ from datetime import datetime
 from errno import EPIPE
 import fcntl
 from functools import wraps
-from os import _exit, environ, kill
+from os import _exit, environ, getpid, kill
+from os.path import join
 from select import select
 from signal import signal, SIG_DFL, SIGINT, SIGTERM
 import struct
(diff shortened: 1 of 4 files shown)


bundlewrap/utils/ui.py
```diff
@@ -167,14 +193,19 @@ def deactivate(self):
...
     @clear_formatting
     @add_debug_indicator
+    @capture_for_debug_logfile
     @add_debug_timestamp
     def debug(self, msg, append_newline=True):
         if self.debug_mode:
             with self.lock:
                 self._write(msg, append_newline=append_newline)
+        if self.debug_log_file:
+            self.debug_log_file
```
vain: What's this statement for?

trehn: Whoopsie.

vain:
Oh, yeah, output files are rather ... large. One run of `bw verify` on one node produced 440kb. I haven't tried our full repo yet.

Output format should probably be revised to make this more readable. Off the top of my head: The node name should always came right after `[DEBUG]`.

trehn:
The worker pool stuff accounts for about one third of all lines, we can probably get rid of that.

trehn:
As for output format, we should probably be more deliberate about *what* we're logging, too. But that's another issue IMO.

PR #297, opened 2016-12-18
Improve psql

vain:
- Fix crash when changing DB ownership
- Don't insist on defining passwords when a role is to be deleted


trehn:
thanks!

PR #308, opened 2017-01-03
Update pkg_pacman for recently addded package cache

vain:
I'm not too happy with that `tarball` stuff. It feels like a hack. Maybe we should remove this in bw 3.0 and advise people to set up their own package repos.


trehn:
thanks!

Issue #310, opened 2017-01-06
env var `BW_REPO_PATH` to override CWD for CLI


trehn:
would help with vagrant et al

PR #313, opened 2017-01-15
add `BW_REPO_PATH` and `bw —repository`

trehn:
@vain I'm not sure if we need both, but I can't really settle for just one or the other...


trehn:
CC #310

vain:
I think it's just fine to have both options available. Sometimes it's easier to set env vars, sometimes it's easier to change argv ...

Did some test runs, works fine.

Me, personally, I'd either call it `BW_REPOSITORY` or `bw --repo-path` for the sake of consistency.

PR #316, opened 2017-01-18
add `—state`, item and attribute args to `bw items`

trehn:
This allows for exploration of item cdicts and sdicts from the CLI. That has previously only been possible for cdicts through `bw hash -d`.

### DIFF
--- a/bundlewrap/cmdline/items.py
+++ b/bundlewrap/cmdline/items.py
@@ -6,7 +6,8 @@ from os.path import dirname, exists, join
 from sys import exit
 
 from ..utils.cmdline import get_node
-from ..utils.text import force_text, mark_for_translation as _
+from ..utils.statedict import statedict_to_json
+from ..utils.text import force_text, mark_for_translation as _, red
 from ..utils.ui import io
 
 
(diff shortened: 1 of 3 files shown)


vain:
Output can be a little misleading. Consider a user which is to be deleted:

    users = {
        'hans': {
            'delete': True,
        },
    }

The user item will return `None` as cdict:

    def cdict(self):
        if self.attributes['delete']:
            return None

Thus, you only get an empty line as output. Or you get a `'NoneType' object is not subscriptable` exception when trying to retrieve one single attribute.

Maybe that would be sufficient:

```diff
diff --git a/bundlewrap/cmdline/items.py b/bundlewrap/cmdline/items.py
index a572b99..995a2c0 100644
--- a/bundlewrap/cmdline/items.py
+++ b/bundlewrap/cmdline/items.py
@@ -72,10 +72,13 @@ def bw_items(repo, args):
             statedict = item.sdict()
         else:
             statedict = item.cdict()
-        if args['attr']:
-            io.stdout(repr(statedict[args['attr']]))
+        if statedict is None:
+            io.stdout(_("<item cdict/sdict is None>"))
         else:
-            io.stdout(statedict_to_json(statedict, pretty=True))
+            if args['attr']:
+                io.stdout(repr(statedict[args['attr']]))
+            else:
+                io.stdout(statedict_to_json(statedict, pretty=True))
     else:
         for item in sorted(node.items):
             if args['show_repr']:
```

Still a little bit ambiguous, though, so maybe set the exit code to something > 1. (I'd expect 1 if the item or node does not exist.)

vain:
Jup, that’s better. :-)

Issue #136, opened 2014-09-16
RUNLEVEL=1 for pkg_apt installs


trehn:
`RUNLEVEL=1 aptitude install $daemon` prevents the daemon from being started after installing

bkendinibilir:
@trehn +1, I need this for automatic provisioning of a vagrant-based development vm.
It would be nice, if there is an option for apt_pkt to set this individually.

Issue #307, opened 2017-01-02
[RFC] creation-only item attributes


trehn:
So far we have religiously avoided attributes on items that can only be enforced when the item is created on the node, but cannot be changed afterwards. The reasoning being that it may not be apparent to the user, which attributes can be fixed by `bw apply` when an item already exists. I do not think documentation alone can solve this.

Here are some examples that would benefit from a solution:

* #306 
* block devices for ZFS pools
* encoding for postgres databases
* ...

The best idea I have right now is to add an explicitly named subdict:

```python
postgres_dbs = {
    "mydb": {
        '__enforce_when_creating__': {
            'encoding': "utf-8",
        },
        'owner': "root",
    },
}
```

Everything below `__enforce_when_creating__` cannot be part of the item's `cdict()`.

PR #319, opened 2017-01-18
Pkg apt runlevel 1

vain:
CC #136


trehn:
another case of #307 

I'd prefer to wait until we have a solution there

trehn:
merged your changes into #329

Issue #208, opened 2016-02-19
`bw test` warning for orphaned bundles


trehn:
should display unused bundles as warning

PR #320, opened 2017-01-18
Warn about unused bundles

vain:
CC #208

### DIFF
--- a/bundlewrap/cmdline/test.py
+++ b/bundlewrap/cmdline/test.py
@@ -8,7 +8,7 @@ from ..concurrency import WorkerPool
 from ..plugins import PluginManager
 from ..repo import Repository
 from ..utils.cmdline import get_target_nodes
-from ..utils.text import bold, green, mark_for_translation as _, red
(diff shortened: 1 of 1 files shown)


trehn:
I think it would be better to put this towards the end of the output, so it doesn't scroll by too quickly...

vain:
Fun fact: I deliberately put it at the beginning because it's just a set of mildly important warnings and it's very fine if it scrolls by. %)

I'm fine with both, though. Will change.

vain:
Thinking about it ... it would be nice to have a quick way to check for orphaned bundles without having to test *everything*. (This was a nice side effect of my initial version.)

Maybe move the output for orphaned bundles to the end and add a command line switch `--only-orphaned-bundles` which omits all other tests?

trehn:
oh and please add a TODO for making this an error in some way (maybe with a CLI switch) in 3.0

trehn:
nah, just leave it at the beginning then

trehn:
I think in 3.0 we need to come up with a more flexible way of specifiying which tests should be run.

trehn:
LGTM, waiting for build

PR #322, opened 2017-01-21
node.py: move oralelinux to OS_FAMILY_REDHAT

rullmann:
Oracle Linux is actually based on RHEL. The changes are minimal and full compatibility should be given when it comes to bw.


trehn:
thx

PR #323, opened 2017-01-28
Deal with SUID and SGID on directories

vain:


trehn:
thank you!

vain:
Oh, just for the record.

The syscall itself does "the right thing" on Linux:

```c
#include <stdio.h>
#include <stdlib.h>
#include <sys/stat.h>

int
main(int argc, char **argv)
{
    long int mode;

    if (argc == 3)
    {
        mode = strtol(argv[1], NULL, 8);
        if (chmod(argv[2], mode) == -1)
            perror("chmod");
    }

    return 0;
}
```

`./foo 0755 testdir` clears the upper bits.

Busybox's `chmod` also behaves this way.

GNU coreutils' `chmod` clears the bits if you invoke it as `chmod =0755 testdir`. Don't get excited, this syntax is not understood by any other tool (at least not FreeBSD, NetBSD, OpenBSD, or Busybox).

Why does GNU coreutils do this, anyway?

There's this bug report:

https://debbugs.gnu.org/cgi/bugreport.cgi?bug=8391

Quoting Eric Blake:

> we have chosen to define that they are not cleared (accidentally removing the sticky bit of a directory when you only meant to change access permissions is a security hole).

You might wonder, "but I do a `chmod 0755`, so I explicitly set the upper bits to zero, that's not an accident!" No, because people might not understand/know that chmod automatically uses octal numbers, so they might think that they *have* to add a leading zero in order to make it use octal. So the `0` in `0755` might not mean "clear upper bits" but "this is an octal number and I want to set the mode to `755`". No, GNU chmod does not support decimal modes and, yes, the manual states that a numeric argument is interpreted as octal.

To be fair, some routines that convert from decimal to octal automatically add a leading zero, which could indeed cause confusion if you're calling chmod from a script.

Very well. They also discuss in that bug report, that you could do `chmod 00755 testdir` which finally is explicit enough. Lo and behold, this syntax is understood by recent versions of GNU coreutils, FreeBSD, NetBSD, OpenBSD, and Busybox.

But then again ...

See how that bug report is from 2011? Yeah, you guessed right. Ubuntu 12.04 comes with coreutils 8.13 which does not yet support `chmod 00755 testdir`. You need Ubuntu 14.04 or newer. CentOS 7 is fine, though, but I haven't checked older CentOS releases. XenServer 6.5 is fine, too, but it comes with an extremely old version of chmod, so maybe that's patched by Citrix.

Is it really patched? See this discussion:

http://lists.gnu.org/archive/html/bug-coreutils/2006-07/msg00124.html

GNU chmod *used to clear the upper bits*, but they changed that behaviour in 2006. Now consider this output from a XenServer 6.5 system:

```
# chmod --version
chmod (GNU coreutils) 5.97
Copyright (C) 2006 Free Software Foundation, Inc.
```

That chmod might be *old enough* to still show the unchanged behaviour ...

To wrap things up, invoking chmod with an extra leading `0` would have been nice, but I think we should keep the current version.

PR #325, opened 2017-02-01
Tables

trehn:


vain:
$ bw stats
	  File "/home/void/git/bundlewrap/bundlewrap/cmdline/__init__.py", line 67, in wrapper
		return f(*args, **kwargs)
	  File "/home/void/git/bundlewrap/bundlewrap/cmdline/__init__.py", line 108, in main
		io.debug(_("invocation: {}").format(" ".join([force_text(arg) for arg in argv])))
	  File "/home/void/git/bundlewrap/bundlewrap/utils/ui.py", line 70, in wrapped
		return f(self, msg, **kwargs)
	  File "/home/void/git/bundlewrap/bundlewrap/utils/ui.py", line 37, in wrapped
		return f(self, "[DEBUG] " + msg, **kwargs)
	  File "/home/void/git/bundlewrap/bundlewrap/utils/ui.py", line 56, in wrapped
		ansi_clean.sub("", msg).rstrip("\n") + "\n"
	AttributeError: 'function' object has no attribute 'sub'

trehn:
fixed

bundlewrap/utils/ui.py
```diff
@@ -96,7 +96,7 @@ def term_width():
 def page_lines(lines):
     lines = list(lines)
     line_width = max([len(line) for line in lines])
-    if line_width > term_width():
+    if TTY and line_width > term_width():
```
vain: Ideally, this would use the raw line width, not the one including escape sequences. Doesn't break things but it's a little confusing.

trehn:
Not in this PR, but I'm pondering an env var `BW_TABLE_STYLE` that defaults to the current tables, but can be set to `ascii` for tables made entirely of `+-|` instead of the fancy Unicode characters (which don't line up well in every font). And maybe a third option `grep` for a more CSV-like format.

trehn:
Try `bw nodes -a`, it's awesome.

PR #327, opened 2017-02-21
systemd: Support 'None' as value for 'running'

vain:
We often have situations where we don't want bw to start or stop a
daemon. Maybe enable it, but don't force start/stop.

### DIFF
--- a/bundlewrap/items/svc_systemd.py
+++ b/bundlewrap/items/svc_systemd.py
@@ -100,10 +100,7 @@ class SvcSystemd(Item):
     @classmethod
     def validate_attributes(cls, bundle, item_id, attributes):
         for attribute in ('enabled', 'running'):
-            if (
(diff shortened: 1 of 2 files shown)


bundlewrap/items/svc_systemd.py
```diff
@@ -100,9 +100,12 @@ def sdict(self):
     @classmethod
     def validate_attributes(cls, bundle, item_id, attributes):
         for attribute in ('enabled', 'running'):
-            if not isinstance(attributes.get(attribute, True), bool):
+            if (
+                not isinstance(attributes.get(attribute, True), bool) and
+                not attributes.get(attribute, None) is None
+            ):
```
trehn: `if attributes.get(attribute, None) not in (True, False, None)`

trehn:
thx

PR #328, opened 2017-03-06
Cache packages per-node and cache full IDs

vain:
It appears this was a *global* cache, covering all nodes. Whoops.

### DIFF
--- a/bundlewrap/items/pkg.py
+++ b/bundlewrap/items/pkg.py
@@ -28,7 +28,7 @@ class Pkg(Item):
 
     def fix(self, status):
         try:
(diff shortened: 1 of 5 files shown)


bundlewrap/items/pkg.py
```diff
@@ -28,7 +28,7 @@ def __repr__(self):
 
     def fix(self, status):
         try:
-            self._pkg_install_cache.remove(self.name)
+            self._pkg_install_cache.get(self.node.name, {}).remove(self.id)
```
trehn: shouldn't this be

```python
self._pkg_install_cache.get(self.node.name, set()).remove(self.id)
```

trehn:
thx

PR #331, opened 2017-04-06
added dash docset build script

sHorst:


trehn:
Hi, thank you for opening this PR.

I like the idea of having a Dash docset for BundleWrap, but I'm not convinced this has to be in the main repository. Aside from the icons, there seems to be nothing here that's really specific to BundleWrap. It seems you did not write these scripts yourself, so the licensing is unclear as well.

I hate turning down PRs, but I really believe your work will stand better on its own as a generic mkdocs->docset converter - with just some extra parameters for BundleWrap.

PR #334, opened 2017-04-12
Exclude dummy nodes from `bw lock`

vain:
CC #333.


trehn:
This introduces some asymmetry. `bw apply` and `bw verify` will tell you that a dummy node "has no items", while this just silently skips dummy nodes.

vain:
I've pushed an alternative fac^W implementation that handles dummy nodes in `lock.py`.

This feels a little convoluted, though. IMHO, it would be easier to have a little function in `cmdline/lock.py` that filters all dummy nodes and prints that `has no items` line. This function would be called from `bw_lock_add` and friends. How about that?

trehn:
👍

trehn:
thx

PR #338, opened 2017-05-05
postgres_db: Allow customization of encoding and collation

vain:
It's not possible to change these attributes after the DB has been created, so they're all creation-time only. :/


trehn:
thanks!

PR #339, opened 2017-05-15
make copy of values if we want to change them for diff

sHorst:
There is a problem with the validation of dict values in items. BW will add an empty item to the end. I narrowed it down to diff_value_list. If you make a copy there before you add items to the list, all is working correct.

This problem only occures, in the interactive mode, since the sdict function is called twice, and the second time the sdict alreay has the empty element, and therefor the validation fails, which states fix failed, but it happened and worked.


trehn:
Thanks, but I already fixed that in ba233177b3c209750a156e812b56416853f82eb5 (see IRC).

PR #342, opened 2017-05-22
fix issue#341

sHorst:
Pull request for issue#341


trehn:
Thank you! There were a couple of problems though:

*  it's "groups", not "group"
* you can't use `self.attributes.get()` here since defaults have already been set when `get_auto_deps()` is called
* thus you end up with a `TypeError: argument of type 'NoneType' is not iterable` when no groups are set

I went ahead and fixed the issue you reported in eb80aa883d8bbf5cb5df53b1e8229e52dda62917.

PR #346, opened 2017-06-13
Actions: Pipes

vain:
```
actions = {
    'decrypt': {
        'command': 'cryptsetup --batch-mode luksOpen /dev/foo',
        'data_stdin': 'my_secret_key',
    },
}
```

Should the value of `data_stdin` be shown to the user? I don't think so. It could be large or binary. If we show it, we must also crop it if needed.

### DIFF
--- a/bundlewrap/items/actions.py
+++ b/bundlewrap/items/actions.py
@@ -69,9 +69,10 @@ class Action(Item):
                 ))
                 return (self.STATUS_SKIPPED, ["unless"])
 
-        question_body = self.attributes['command']
+        question_body = ""
         if self.attributes['data_stdin'] is not None:
-            question_body += "\n\n<data being fed to stdin>"
+            question_body += "<" + _("data") + "> | "
+        question_body += self.attributes['command']
         if self.comment:
             question_body += format_comment(self.comment)
 
--- a/bundlewrap/items/actions.py
+++ b/bundlewrap/items/actions.py
@@ -5,6 +5,7 @@ from datetime import datetime
 
 from bundlewrap.exceptions import ActionFailure, BundleError
 from bundlewrap.items import format_comment, Item
+from bundlewrap.utils import Fault
 from bundlewrap.utils.ui import io
 from bundlewrap.utils.text import mark_for_translation as _
 from bundlewrap.utils.text import blue, bold, wrap_question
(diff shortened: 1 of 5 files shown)


bundlewrap/items/actions.py
```diff
@@ -128,13 +131,23 @@ def get_result(self, *args, **kwargs):
         return status_code
 
     def run(self):
+        if self.attributes['data_stdin'] is not None:
+            data_stdin = self.attributes['data_stdin']
+            # Allow users to use either a string/unicode object or raw
+            # bytes.
+            if type(data_stdin) is not bytes:
+                data_stdin = data_stdin.encode('UTF-8')
```
trehn: Should allow Faults as well. I think it would be best to just mirror `bundlewrap.utils.text.force_text`.

bundlewrap/items/actions.py
```diff
@@ -69,6 +70,8 @@ def _get_result(
                 return (self.STATUS_SKIPPED, ["unless"])
 
         question_body = self.attributes['command']
+        if self.attributes['data_stdin'] is not None:
+            question_body += "\n\n<data being fed to stdin>"
```
trehn: Missing _() for translation. Also I think "(data is being fed to stdin)" would be slightly more clear since angle brackets usually indicated some form of placeholder. Or just integrate it into the command display:
```
<data> | my command
```

tests/integration/bw_apply_actions.py
```diff
@@ -46,3 +46,27 @@ def test_action_fail(tmpdir):
...
+
+def test_action_pipe(tmpdir):
+    make_repo(
+        tmpdir,
+        bundles={
+            "test": {
+                'actions': {
+                    "pipe": {
+                        'command': "cat",
+                        'data_stdin': "hello\n",
```
trehn: Maybe sprinkle some umlauts here to test UTF-8 encoding...

bundlewrap/items/actions.py
```diff
@@ -128,13 +133,25 @@ def get_result(self, *args, **kwargs):
         return status_code
 
     def run(self):
+        if self.attributes['data_stdin'] is not None:
+            data_stdin = self.attributes['data_stdin']
+            # Allow users to use either a string/unicode object or raw
+            # bytes -- or Faults.
+            if isinstance(data_stdin, Fault):
+                data_stdin = data_stdin.value
+            elif type(data_stdin) is not bytes:
```
trehn: s/elif/if/

Faults are likely to be Unicode

vain: Yeah.

PR #365, opened 2017-09-21
pkg_openbsd: Fix handling of empty output from "pkg_info"

vain:
If no packages are installed on the system, then "pkg_info" prints
nothing, i.e. an empty string. That's handled badly by split("\n"):

    Python 3.6.2 (default, Jul 20 2017, 03:52:27)
    [GCC 7.1.1 20170630] on linux
    Type "help", "copyright", "credits" or "license" for more information.
    >>> "".split("\n")
    ['']

The following code does not expect to work on an empty string.

splitlines() handles the situation much better by returning an empty
list:

    Python 3.6.2 (default, Jul 20 2017, 03:52:27)
    [GCC 7.1.1 20170630] on linux
    Type "help", "copyright", "credits" or "license" for more information.
    >>> "".splitlines()
    []

### DIFF
--- a/bundlewrap/items/pkg_openbsd.py
+++ b/bundlewrap/items/pkg_openbsd.py
@@ -22,7 +22,7 @@ def pkg_installed(node, pkgname):
         "pkg_info | cut -f 1 -d ' '",
         may_fail=True,
     )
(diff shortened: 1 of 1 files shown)


trehn:
Thanks!

PR #368, opened 2017-09-24
[WIP] try getting multiline job info to work

trehn:
I've not yet given up on getting a multiline progress bar to work.

With the changes in this PR, this works as expected (printing an inverted and regular version of the same status line for now):

```python
from time import sleep
from bundlewrap.utils.ui import io


io.activate()
io.stdout("hi")
with io.job("imajob"):
    for i in range(10):
        io.stdout(str(i))
        sleep(0.5)
io.stdout("bye")
io.deactivate()
```

"Real" bw stilll messes up my terminal though.


trehn:
nah

PR #369, opened 2017-09-24
change `Item.BLOCK_CONCURRENT` into a class method

### LINKED ISSUE #367: OpenBSD: bw must not try to create users/groups in parallel
Yargh.

`nodes.py`:
```python
nodes = {
    'node': {
        'bundles': ['bunch-of-users'],
        'cmd_wrapper_outer': 'sh -c {}',
        'hostname': 'overqemu10',
        'os': 'openbsd',
        'os_version': (6, 1),
    },
}
```

`bunch-of-users/items.py`:
```python
users = {}

for i in range(100):
    users['foobar{}'.format(i)] = {}
```

Result: Some users are created, others are not.

Reason:
```
useradd: can't lock `/etc/master.passwd': Resource temporarily unavailable
```

Worse, this leaves the system in a very weird state:

```
# id foobar6 
uid=1002(foobar6) gid=1002(foobar6) groups=1002(foobar6)
# grep foobar /etc/master.passwd                                                       
foobar9:*:1000:1000::0:0::/home/foobar9:/bin/ksh
foobar0:*:1003:1003::0:0::/home/foobar0:/bin/ksh
# grep foobar /etc/passwd         
foobar9:*:1000:1000::/home/foobar9:/bin/ksh
foobar0:*:1003:1003::/home/foobar0:/bin/ksh
```

So, the user "exists", but it doesn't have an entry in those files? The user also still exists after a reboot and it has been successfully written to `/etc/pwd.db`, apparently:

```
# strings /etc/pwd.db | grep foobar6
foobar6
/home/foobar6
foobar6
/home/foobar6
1foobar6_syspatch
foobar6
/home/foobar6
```

This means that you can't even do multiple runs of `bw apply ...` as a workaround, because bw only reads `/etc/master.passwd` to determine if a user exists. Subsequent calls to `useradd` will fail in a corrupted system.

I can reproduce the corruption/inconsistency by running `for i in $(jot 99); do useradd foobar$i & done`. bw is not the root cause of the corruption here, but that's not the point. It must not try to create users in parallel on OpenBSD due to the lock file.

(I haven't been able to reproduce the inconsistency in OpenBSD 5.9, but there's a lock file as well.)

trehn:
fixes #367

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ unreleased
 * removed `bw apply --profiling` (BACKWARDS INCOMPATIBLE)
 * removed `Item.display_keys()` (BACKWARDS INCOMPATIBLE)
 * changed return value of `Item.display_dicts()` (BACKWARDS INCOMPATIBLE)
+* changed `Item.BLOCK_CONCURRENT` into a class method (BACKWARDS INCOMPATIBLE)
 * removed `repo.vault.format()` (BACKWARDS INCOMPATIBLE)
 * removed env vars: BWADDHOSTKEYS, BWCOLORS, BWITEMWORKERS, BWNODEWORKERS (BACKWARDS INCOMPATIBLE)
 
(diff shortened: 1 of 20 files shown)


vain:
:+1:

Pushed an actual fix (438a96a) and some cosmetics.

trehn:
derp, thanks

PR #374, opened 2017-10-05
Add missing calls to io.deactivate()

vain:
For example, running "bw nodes" from a directory which is not a bw repo
would leave the cursor invisible.


trehn:
Thanks, just made sure it would do that for *any* Exception.

PR #375, opened 2017-10-18
fix prepare_dependencies in bw plot

chenull:
The error was:

```
  File "/Users/ayik/Repo/Python/Utils/bundlewrap/bundlewrap/cmdline/plot.py", line 32, in bw_plot_node
    prepare_dependencies(node.items, node.os, node.os_version),
  File "/Users/ayik/Repo/Python/Utils/bundlewrap/bundlewrap/utils/ui.py", line 359, in inner_wrapper
    return wrapped_function(*args, **kwargs)
TypeError: prepare_dependencies() missing 1 required positional argument: 'node_os_version'
```

According to Contributing Doc about CAA, i think this patch is trivial / small so that i don't need to sign CAA.

### DIFF
--- a/bundlewrap/cmdline/plot.py
+++ b/bundlewrap/cmdline/plot.py
@@ -29,7 +29,7 @@ def bw_plot_node(repo, args):
     node = get_node(repo, args['node'], adhoc_nodes=args['adhoc_nodes'])
     for line in graph_for_items(
         node.name,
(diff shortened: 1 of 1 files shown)


trehn:
Thank you!

trehn:
Your fixed has now been released as part of 3.1.1.

PR #379, opened 2017-11-26
pkg_openbsd: Support flavors

vain:
Setting "version" to select a flavor is a bit flawed because you have
to, well, select a *version*. This is not what you actually want most of
the time (at least I don't). I want to be able to select a particular
flavor and, more importantly, it shall default to the "normal" flavor if
unset.

Some examples:

    pkg_add foo
        Installs most recent version of package "foo". If there are
        multiple flavors of that package, users will be asked
        interactively to select one of them.

    pkg_add foo--
        Installs most recent version of package "foo" and selects the
        "normal" flavor.

    pkg_add foo--bar
        Installs most recent version of package "foo" and selects the
        "bar" flavor.

    pkg_add foo-1.0.0-bar
        Explicitly installs version "1.0.0" of flavor "bar".

With this patch, bw will do a "pkg_add foo--" by default. If you set a
flavor, "pkg_add foo--bar" will be used. If you set a version, the
flavor attribute will be ignored completely and "pkg_add
foo-version-bar" will be used.

I had to adjust PKGSPEC_REGEX because it was too greedy.

### DIFF
--- a/bundlewrap/items/pkg_openbsd.py
+++ b/bundlewrap/items/pkg_openbsd.py
@@ -99,7 +99,7 @@ class OpenBSDPkg(Item):
         version, flavor = pkg_installed(self.node, self.name)
         return {
             'installed': bool(version),
-            'flavor': flavor if version else _("none"),
+            'flavor': flavor if flavor else _("none"),
             'version': version if version else _("none"),
         }
 
--- a/bundlewrap/items/pkg_openbsd.py
+++ b/bundlewrap/items/pkg_openbsd.py
@@ -9,11 +9,14 @@ from bundlewrap.items import Item
 from bundlewrap.utils.text import mark_for_translation as _
 
 
-PKGSPEC_REGEX = re.compile(r"^(.+)-(\d.+)$")
+PKGSPEC_REGEX = re.compile(r"^([^-]+)-(\d[^-]+)(-(.+))?$")
 
 
-def pkg_install(node, pkgname, version):
-    full_name = "{}-{}".format(pkgname, version) if version else pkgname
+def pkg_install(node, pkgname, flavor, version):
+    if version:
+        full_name = "{}-{}".format(pkgname, version)
+    else:
+        full_name = "{}--{}".format(pkgname, flavor)
     return node.run("pkg_add -r -I {}".format(full_name), may_fail=True)
 
 
@@ -23,10 +26,15 @@ def pkg_installed(node, pkgname):
         may_fail=True,
     )
     for line in result.stdout.decode('utf-8').strip().splitlines():
-        installed_package, installed_version = PKGSPEC_REGEX.match(line).groups()
+        installed_package, installed_version, _, installed_flavor = \
+            PKGSPEC_REGEX.match(line).groups()
         if installed_package == pkgname:
-            return installed_version
-    return False
+            # If our regex didn't match a flavor, then this is
+            # equivalent to using the "normal" flavor.
+            if installed_flavor is None:
+                installed_flavor = ""
+            return installed_version, installed_flavor
+    return False, None
 
 
 def pkg_remove(node, pkgname):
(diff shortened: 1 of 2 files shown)


vain:
Maybe `version` should be removed in 4.0. Is it even possible in OpenBSD to install older versions of a package? I don't see multiple package versions on mirrors (http://ftp.halifax.rwth-aachen.de/openbsd/6.2/packages/amd64/, for example).

I also get the feeling that `version` was only added to provide the ability to select a flavor.

trehn:
Two things here that rub me wrong:

1) `flavor` and `version` being mutually exclusive. Sticking with the example from the docs, why not this:

```python
pkg_openbsd = {
    'irssi': {
        'flavor': 'socks',
        'version': '1.0.4p0',
    },
}
```

Resulting in `pkg_add irssi-1.0.4p0-socks` and

```python
pkg_openbsd = {
    'irssi': {
        'version': '1.0.4p0',
    },
}
```

resulting in `pkg_add irssi-1.0.4p0` and

```python
pkg_openbsd = {
    'irssi': {
        'flavor': 'socks',
    },
}
```

Resulting in `pkg_add irssi--socks`. Previously, people might have used

```python
pkg_openbsd = {
    'irssi': {
        'version': '-socks',
    },
}
```

To just get the flavor without an explicit version, but that would still work.

2) Changing the return type of `pkg_installed()`. It happens to work, but `bundlewrap.items.pkg.Pkg` might expect this to return just a boolean in the future. It also creates asymmetry between `pkg_openbsd` and `pkg_pip`, where I didn't use the `Pkg` class because I needed to handle versions there too.


From reading `man pkg_add`, it appears to me as if the ability to specify a version is probably only relevant when installing from local tarballs.

And then there's this can of worms, which we don't handle at all yet:

```
There is also an ambiguity related to ports with multiple branches. For instance ‘pkg_add python’ is
 ambiguous, as there are several versions of python in the ports tree. So is ‘pkg_add postfix’. The
 special form ‘pkgname%branch’ can be used to restrict matches to a branch matching the
 pkgpath(7).

The above ambiguities can be resolved using ‘pkg_add postfix%stable’ and ‘pkg_add python%3.4’, respectively.

All paths recognize certain special sequences, which are expanded as follows:
[...]
```

vain:
1.  Yeah, alright. Tried to fix that. `flavor` and `version` can be used together now. Sorry for the `if` mess. :( I might be braindead.
2.  I'm confused. `pkg_openbsd` works similar to `pkg_pip`, i.e. `pkg_installed()` is not a method but a module level function. (I totally agree that these mixed return types are not nice, but that applies to `pkg_pip` as well.)
3.  Meh. Can we please close our eyes and sing a happy song instead?

bundlewrap/items/pkg_openbsd.py
```diff
@@ -60,12 +88,18 @@ def fix(self, status):
...
+                self.attributes['flavor'],
+                self.attributes['version']
+            )
 
     def sdict(self):
-        version = pkg_installed(self.node, self.name)
+        version, flavor = pkg_installed(self.node, self.name)
         return {
             'installed': bool(version),
+            'flavor': flavor if version else _("none"),
```
trehn: flavor if flavor?

bundlewrap/items/pkg_openbsd.py
```diff
@@ -40,6 +65,7 @@ class OpenBSDPkg(Item):
     BUNDLE_ATTRIBUTE_NAME = "pkg_openbsd"
     ITEM_ATTRIBUTES = {
         'installed': True,
+        'flavor': "",
```
trehn: Shouldn't flavor default to `None`?

>flavor = None,   version = "-foo":  "pkgname--foo"  (backwards compat)

bundlewrap/items/pkg_openbsd.py
vain: Yes, will fix, thanks.

bundlewrap/items/pkg_openbsd.py
vain: I think `flavor` should to default to `""`, since I believe that this is what you want most of the time.

It will be a change in behavior, yes. You might call it a "fix", because bw previously simply failed to install a package -- unless you set `version` to `-foo`, which is neither obvious nor documented.

If you prefer to have it default to `None`, I'd like to add a TODO item to change it to `""` for bw 4.0.

trehn:
Thanks!

PR #381, opened 2017-12-19
honour BW_SSH_ARGS for scp command

chenull:
Hi, when using custom SSH config defined in `BW_SSH_ARGS`, `scp` command may be failed since it's still using default ssh config file. This patch will make `scp` command honour `BW_SSH_ARGS`


trehn:
Thank you!

PR #388, opened 2018-02-04
added sha256

sHorst:
- added sha256 sum to PATH_INFO.


trehn:
Can you elaborate on why you think this is required and where it should be used?

SHA1 is a deliberate choice for its speed advantage.

sHorst:
i try to implement a download and verify plugin, which downloads a url and verifies, that it is correct. The SHA1 is deprecated because it starts to degrate from a security perspective. It is good to check, if an upload was correct, but not to verify that an untrusted 3rd party delivered the correct file. That is why i want to check the sha256 and sha1 is not secure enough.

trehn:
OK, thanks.

PR #392, opened 2018-03-19
postgres_*: No longer throw exceptions

vain:
Assigning postgres_* items to a node and then running "bw verify $node"
results in some RemoteExceptions. That's because we try to run "sudo
postgres ...", but the user "postgres" does not yet exist. Usually,
bundles will have dependencies for that (install pkg_apt:postgresql
before creating any databases), but "bw verify" ignores dependencies.

So, allow these ".run()" calls to fail. This makes "bw verify" report
items as failed -- this is what we want.


trehn:
Thanks!

PR #401, opened 2018-05-29
Generate better Error-Message for Null-Key Dict-Access in Mako-Templates

MaZderMind:


bundlewrap/items/files.py
```diff
@@ -105,6 +105,8 @@ def content_processor_mako(item):
             # error message - even though we can't pinpoint the excat
             # location of the error. :/
             e = _("Undefined variable (look for '${...}')")
+        elif isinstance(e, KeyError) and str(e) == "None":
```
trehn: What's so special about None as a key? Why not do it for all keys:

```python
elif isinstance(e, KeyError):
    e = _("KeyError: {}").format(str(e))
```

MaZderMind: true, true. Force-Pushed an update.

PR #403, opened 2018-07-12
made code in docu path independent

sHorst:
The code which is shown in the documentation for searching a nodes folder for nodes files is not path independent. If you try to use this code with an script it does not work as expected, since it does not find its nodes files.

This patch adds the repo path to the mix, which fixes the problem. Additionaly it adds the os.path.join method, which makes this code run on windows as well.


trehn:
thanks!

PR #413, opened 2018-12-23
Fix an edge case when trying to create a symlink that ends with 'directory'

schuppentier:
As the end of the output of `file -bh` is "user controlled" for symlinks, the beginning of the string should be used to avoid problems.

In my case, this bug was triggered when creating a vhost configuration for FusionDirectory.

### DIFF
--- a/bundlewrap/utils/remote.py
+++ b/bundlewrap/utils/remote.py
@@ -12,14 +12,14 @@ def _parse_file_output(file_output):
     if file_output.startswith("cannot open "):
         # required for Mac OS X, OpenBSD, and CentOS/RHEL
         return ('nonexistent', "")
+    elif file_output.startswith("symbolic link to ") or \
+            file_output.startswith("broken symbolic link to "):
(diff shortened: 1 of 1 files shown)


trehn:
Thanks for reporting this, but this fix breaks another case:

```
$ mkdir test
$ chmod +s test
$ file -bh test
setuid, setgid, directory
```

So I guess the better fix would be to move the `if` for symlink up before the `if` for directories. Do you want to do that?

schuppentier:
I actually did that at first, but thought, the other change was cleaner. Maybe it would be better to completely switch from `file` to `stat` and include this in the call to `stat`?

trehn:
Also, Merry Christmas. 🎄

trehn:
Ah, with stat there is no way to tell the difference between text and non-text files. It might still be worth it as an optimization to call `file` only on "regular files" as identified by `stat`. But that's a different issue.

Thank you!

PR #416, opened 2019-01-31
Fixed  #396

### LINKED ISSUE #396: Duplicate item definition if two bundles need the same {apt,pacman,pip}-package
If two bundles require the same apt (or any other) package and the respective `pkg_apt` items are added to the bundles, bw throws a BundleError with "duplicate definition". While this is useful to make sure no two bundles define different attributes for files or directories, I don't see the point in throwing the Error when both items' definitions are exactly the same. This will probably be the case with dependencies to packages, as they are usually just installed without any other metadata.

Am I just getting the wrong idea here and there is a better approach to such situations? Or should I create a dedicated bundle with only the `apt_pkg` dependency that both bundles should depend on?

DasLampe:


trehn:
Um, no. All you're doing here is suppressing the error message. You are not addressing any of the reasons why `bw` is calling this an error:

* Conflicting item defintions. You might end up installing and the uninstalling the same package during a single run of `bw apply`. That should be an error.
* Redundant checking of the same item. Even if multiple instances of the package item have the same configuration, you end up checking the same item over and over again.
* Dependencies. You might be lucky in that everything depending on a package might end up depending on the first instance of the package. This can work, but it's pure luck. Don't rely on it.

PR #421, opened 2019-02-25
Fault: Support .b64encode()

vain:
We have *a lot* of these in our k8s stuff:

```python
k8s_secrets = {
    'foo': {
        'context': {
            'username': b64encode(str(teamvault.username('1234')).encode('UTF-8')).decode('UTF-8'),
        },
        'manifest_file': 'foo.yaml',
        'manifest_processor': 'mako',
    },
}
```

That's wrong because Faults will get resolved too early … so, that b64encode dance must be done in the template. That's pretty noisy and a little annoying. :)

With this PR, you can do the following and be done with it:

```python
k8s_secrets = {
    'foo': {
        'context': {
            'username': teamvault.username('1234').b64encode(),
        },
        'manifest_file': 'foo.yaml',
        'manifest_processor': 'mako',
    },
}
```

### DIFF
--- a/bundlewrap/utils/__init__.py
+++ b/bundlewrap/utils/__init__.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
+from base64 import b64encode
 from codecs import getwriter
 from contextlib import contextmanager
 import hashlib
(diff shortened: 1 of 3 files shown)


trehn:
Thanks!

PR #437, opened 2019-11-05
custom SSH arguments per-node

trehn:


vain:
Why did we need this again?

trehn:
We didn't, that's why I never merged it. It's just something that people occasionally ask for, but so far it's never really been a good solution for anything. Closing now due to merge conflicts, we can dig it back up if we do need it after all someday.

PR #439, opened 2019-11-12
k8s raw collisions

trehn:


bundlewrap/items/kubernetes.py
```diff
@@ -242,34 +258,7 @@ def get_auto_deps(self, items):
...
-                bundle=self.bundle.name,
-                node=self.bundle.node.name,
-                regex=self.NAME_REGEX,
-            ))
-        else:
-            return name
+        return self.name.split("/", 2)[1]
 
     @property
     def resource_name(self):
```
vain: This method should be removed, too. Otherwise, we don’t check the actual k8s object name but just it’s bw name.

vain:
Works as intended, let’s merge this.

PR #440, opened 2019-12-24
metaproc flowcontrol

CroneKorkN:


bundlewrap/repo.py
```diff
@@ -541,88 +541,131 @@ def _build_node_metadata(self, blame=False):
...
+
+            print('populate_dependencies')
+            
+            # collect all metaprocs
+            metaprocs = {}
             for node_name in list(self._node_metadata_partial):
+                node = self.get_node(node_name)
+                for name, func in node.metadata_processors:
+                    func.__setattr__('__node', node)
+                    metaprocs[name] = func
```
trehn: You're mixing things up here, overwriting metadata processors with the same on different nodes. Get rid of the setattr and use a two-tuple like `blacklisted_metaprocs`.

bundlewrap/repo.py
```diff
@@ -541,88 +541,131 @@ def _build_node_metadata(self, blame=False):
...
+            
+            # dependecy assignment helper
+            import re
+            def populate_dependency(name, after):
+                for metadata_processor_name, metadata_processor in metaprocs.items():
+                    if re.match(name, metadata_processor_name):
+                        metadata_processor.__setattr__('__after', [
+                            after, 
+                            *getattr(metadata_processor, '__after')
+                        ])
```
trehn: metadata_processor.__after.append(after)

bundlewrap/repo.py
```diff
@@ -541,88 +541,131 @@ def _build_node_metadata(self, blame=False):
...
+            
+            # assign dependencies
+            for metadata_processor_name, metadata_processor in metaprocs.items():
+                for name in getattr(metadata_processor, '__before'):
```
trehn: why are you using getattr like this? `metadata_processor.__before`

bundlewrap/repo.py
```diff
@@ -541,88 +541,131 @@ def _build_node_metadata(self, blame=False):
...
+                    continue
+                
+                # check dependencies
+                dependency_missing = False
+                after = getattr(metadata_processor, '__after')
+                if after:
+                    print('####### {} after {}'.format(metadata_processor_name, after))
+                    for dependency in after:
+                        for name, func in metaprocs.items():
+                            if re.match(name, dependency):
```
trehn: have you measured the performance impact of all that regex matching?

bundlewrap/repo.py
```diff
@@ -541,88 +541,131 @@ def _build_node_metadata(self, blame=False):
...
+                                dependency_missing = True
```
trehn: I don't get it. You're just checking if the dependency is present in the list of metadata processors, but aren't tracking it it has actually run yet?

.gitignore
```diff
@@ -1,2 +1,7 @@
 /dist/
 /docs/build/
+
+bundlewrap/__pycache__/
+bundlewrap/cmdline/__pycache__/
+bundlewrap/items/__pycache__/
+bundlewrap/utils/__pycache__/
```
trehn: just add __pycache__ to your global gitignore

bundlewrap/repo.py
CroneKorkN: `AttributeError: 'function' object has no attribute '_Repository__after'`

bundlewrap/repo.py
CroneKorkN: Its neglectable; about 30 microseconds per node.

CroneKorkN: No, my measurement wasnt meaningfull. Will test after further discussion.

trehn:
obsoleted by metastacks

PR #447, opened 2020-01-03
Metaprocs: Detect when users return an original metadata object with an invalid flag

vain:
This is illegal because it can result in duplicate metadata, like
merging lists in merge_dict().

### DIFF
--- a/bundlewrap/repo.py
+++ b/bundlewrap/repo.py
@@ -577,6 +577,7 @@ class Repository(object):
                             ))
                             raise exc
                         processed_dict, options = check_metadata_processor_result(
+                            input_metadata,
                             processed,
                             node.name,
                             metadata_processor_name,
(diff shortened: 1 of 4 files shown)


bundlewrap/repo.py
```diff
@@ -562,7 +562,10 @@ def _build_node_metadata(self, blame=False):
                             # figuring out what changed...
                             input_metadata = deepcopy_metadata(self._node_metadata_partial[node.name])
                         else:
-                            # ...but we can't always do it for performance reasons.
+                            # ...but we can't always do it for performance reasons
+                            # and because we would like to detect if a metaproc
+                            # returned the original dict + {DEFAULTS,OVERWRITE},
+                            # which is illegal.
```
trehn: I'm a little confused. This new check works with and without the deepcopied metadata, so why the comment?

vain: Yeah, brainfart, sorry.

Issue #383, opened 2018-01-02
Consider storing softlocks under /var/tmp instead of /tmp


vain:
/tmp is cleared on both Linux and the BSDs during boot. When you lock a node and then reboot it, the lock is lost. I don't think many people have that in mind or even know about it.

On the other hand, /var/tmp will be preserved.

OpenBSD 6.2, though: No effect. /var/tmp is a symlink to /tmp.

trehn:
To do this in a backwards-compatible way, we'd have to write and check both locations until 4.0. Not sure if worth it. Happy to change it with 4.0.

vain:
Yes, +1 for 4.0.

PR #448, opened 2020-01-08
[bw 4.0] Move SOFT_LOCK_* to /var/tmp

vain:
CC #383.

Since we're talking about bw 4.0 these days, let's not forget about this. :)


trehn:
obsoleted by #462

Issue #449, opened 2020-01-14
New metadata processor API


trehn:
```python
# this replaces "return d, DONE, DEFAULTS"
@metadata_defaults
def foo():
    return {'bar': 5}

@metadata_processor
def proc(metadata):
    # this little monkeypatched .query() helper makes is easier to write 
    # stuff that doesn't assume any pre-existing metadata
    return {
        'baz': metadata.query('frob/gobbel', 'default'),
    }  # RUN_ME_AGAIN and OVERWRITE are now implied
```

Get rid of `OVERWRITE` entirely, it's too niche and can be accomplished with regular metadata_processors.

We could test metadata_processors by passing them an empty dict. They should never rely on anything being in there as that makes it hard to turn stuff that was always included in static metadata into something that can be provided by another metaproc.

It's tempting to replace `.query()` with full-blown JSONPath, but I believe for our purposes it's more useful to just have the ability to easily traverse nested keys and require a mandatory default value.

trehn:
Going further: This replaces the ability of groups to define metadata and bundles. Bundles can only be set at node-level, but bundles can include other bundles. Bundle order matters and determines the merge order of `metadata_defaults`.

TBD: Are groups assembled before or after metadata? Is it more useful to define groups based on metadata or metadata based on groups? I heavily lean towards the former.

trehn:
A node includes bundles A, B and C. Bundles A and C both include bundle D. What is the merge order for `metadata_defaults`? DABC? ADBC? ABDC? ABCD?

trehn:
Do metadata_processor have to be able to alter the original metadata dict? Couldn't they just return a dict that would be merged on top of the original one? I.e. only add/overwrite metadata, never remove.

trehn:
So now I'm expected to write my metaprocs in a way that never assumes any input metadata to be present. In many cases, however, I know that it is an error if the input metadata I want *never* shows up. How can I express that?

```python
@metadata_processor(expect_result=True)
def m(metadata):
    return {}
```

This would throw an error if at the end of processing, this metaproc has never returned anything but an empty dict. Might be helpful to figure out where something went wrong. It's a bit simplistic and binary though. Or we go down the RUN_ME_AGAIN route again and allow metaprocs to return some constant instead of an empty dict that indicates it intends to produce output in the future. This would provide a little more flexibility in deciding whether output will be produced or not.

vain:
To recap:

[...]

PR #451, opened 2020-01-20
[WIP] implement metadata defaults and reactors

trehn:
for #449


bundlewrap/repo.py
```diff
@@ -537,6 +543,28 @@ def _build_node_metadata(self, blame=False):
                             )
                         self._node_metadata_partial[node.name] = new_metadata
 
+            # At this point, static metadata from groups and nodes has been merged.
+            # Next, we look at defaults from metadata.py.
+
+            for node_name in self._node_metadata_partial:
```
vain: I think this is missing a mechanism like `self._node_metadata_static_complete.add(node_name)` to prevent `@metadata_defaults` from being run again and again.

trehn: ✅

PR #459, opened 2020-04-20
k8s: Bump apiVersions

vain:
All apiVersions have been bumped to what's currently listed on
kubernetes.io.

Careful: It appears you can't change the apiVersion of already deployed
objects, so you must re-deploy all affected objects.

### DIFF
--- a/bundlewrap/items/kubernetes.py
+++ b/bundlewrap/items/kubernetes.py
@@ -310,7 +310,7 @@ class KubernetesCronJob(KubernetesItem):
 class KubernetesCustomResourceDefinition(KubernetesItem):
     BUNDLE_ATTRIBUTE_NAME = "k8s_crd"
     KIND = "CustomResourceDefinition"
-    KUBERNETES_APIVERSION = "apiextensions.k8s.io/v1beta1"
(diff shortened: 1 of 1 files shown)


vain:
Might be a good idea to postpone this until bw 4.0 due to the re-deployment pain …

vain:
Should k8s always ignore `apiVersion` … ? Essentially, make it a `when_creating` thing? If we can’t fix it, then why bother checking it?

vain:
I’m going to merge this with #460 and will ignore `apiVersion` just like `status`. Makes it backwards compatible and it’s the right thing to do, IMHO.

vain:
Jesus, this is pointless. k8s decides which apiVersion to use. I just applied this to my cluster:

```
$ cat simple-deploy.yaml 
apiVersion: apps/v1
kind: Deployment
...
```

And what I got back is this:
```
$ cluster-dev -n test get deployment foo -o yaml
apiVersion: extensions/v1beta1
kind: Deployment
...
```

tl;dr:

-   We can't fix apiVersion.
-   We can't even be sure that k8s creates an object with the apiVersion that we specified.

Let's ignore apiVersion except when creating stuff. I'll create a new PR that merges #459 and #460.

PR #461, opened 2020-04-27
k8s: Ignore elusive fields, bump apiVersions

vain:


vain:
This is now backwards compatible and could, in theory, be released in bw 3.x.

trehn:
<3

PR #462, opened 2020-04-28
Flexilocks: Put locks on other nodes

vain:
Solve the problem of unlockable nodes: Hardware, k8s, ...

I'll write docs once we agree that this is the way to go.

Lock format and location changes (would make #448 obsolete). If we need
this sooner than 4.0, we'd have to implement a backwards-compatible
fallback.

### DIFF
--- a/bundlewrap/lock.py
+++ b/bundlewrap/lock.py
@@ -9,7 +9,7 @@ from pipes import quote
 from socket import gethostname
 from time import time
 
-from .exceptions import NodeLockedException, RemoteException
+from .exceptions import NodeLockedException, NoSuchNode, RemoteException
 from .utils import cached_property, tempfile
 from .utils.text import (
     blue,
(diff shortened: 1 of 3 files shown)


bundlewrap/lock.py
```diff
@@ -24,29 +24,7 @@
...
-                "unable to read or parse lock file contents "
-                "(clear it with `bw run {node} 'rm -Rf {path}'`)"
-            ).format(
-                node_bold=bold(node.name),
-                node=node.name,
-                path=HARD_LOCK_PATH,
-                x=red("!"),
-            ))
-            return {}
+LOCK_BASE = "/var/tmp"
```
vain, line 27: Hmm, maybe find a better location. It's fine on Linux, but if you want to use an OpenBSD box as a locking node, then it's bad: It clears /var/tmp on boot.

It's surprisingly hard to find a good and portable path. Maybe `/usr/local/var/bw`? Anything below `/usr/local` should be pretty much guaranteed not to be touched by the OS. Then again, OpenBSD installs packages there. Maybe we should even consider using `/root` ...

vain: (Is using bw without having root access on a node supposed to work?)

trehn:
I like this idea a lot.

vain:
If we accept that

-   this won’t be merged sooner than 4.0,
-   users are required to have root access on managed nodes,

then this PR is good to go.

I suspect that using bw to manage nodes without root access is a very rare case. If we wish to continue supporting this, then I suggest we create another node attribute, `lock_path`, to allow users to overwrite the default of `/var/lib/bw` to something like `/home/tux/.bw_locks`.

vain:
Meh, `/var/lib/bw` might be too ambiguous. Will change to `/var/lib/bundlewrap`.

PR #465, opened 2020-04-29
allow setting group membership at the node

trehn:


trehn:
I am very much inclined to remove members_add and members_remove in 4.0 because of this.

PR #492, opened 2020-05-14
Move repo root detection to constructor of Repository

vain:
This allows you to call "r = Repository('.')" in scripts. Before, you
had to duplicate the root detection in your own code or you had to rely
on using $BW_REPO_PATH.

"Repository()" still yields a repo in /dev/null.


trehn:
lgtm

Issue #488, opened 2020-05-13
k8s_deployments apiVersion?


MichaelKorn:
According to https://docs.bundlewrap.org/items/k8s/ the Deployment apiVersion should be "extensions/v1beta1". But due to #459 that Information is maybe outdated.

Nevertheless, In my opinion it is dangerous to proceed silently to make this change. In my case, bundlewrap stopped working after updating from 3.8.0 to 3.9.0 with GCP. Unfortunately I get the error 'missing required field "selector" in io.k8s.api.apps.v1.DeploymentSpec' with the new API version.

vain:
Hi,

you are right. That was an oversight on my part. Should not have happened, sorry.

To avoid this issue in the future, we’re currently thinking about removing defaults for `apiVersion` in bw 4.0. That would make it mandatory to set `apiVersion` in manifests, meaning bw would no longer guess a version which may or may not be consistent with the rest of your manifest.

As for the docs, I’ll open a PR later with a fix.

PR #501, opened 2020-05-18
k8s_*: No longer provide default for apiVersion

vain:
CC #488.

Providing a default is meaningless, if we don't also validate that the
objects actually comply with that version.


trehn:
lgtm, ignore the Python 2.7 test failures

PR #502, opened 2020-05-18
bring back deepcopy_metadata to replace freeze_object

### LINKED ISSUE #497: allow frozen types in metadata
```
@metadata_reactor
def foo(metadata):
    return {'foo': metadata.get('bar')}
```

If `bar` is a set or dict, we have just returned a frozenset or mappingproxy, which are not allowed in metadata and complicate testing for metadata equality.

Instead of allowing them (and teaching `MetadataJSONEncoder` to handle them), we could also ditch the whole freezing concept and just deepcopy anything we return from `metadata.get()`.

trehn:
fixes #497

### DIFF
--- a/bundlewrap/metadata.py
+++ b/bundlewrap/metadata.py
@@ -1,3 +1,4 @@
+from copy import copy
 from hashlib import sha1
 from json import dumps, JSONEncoder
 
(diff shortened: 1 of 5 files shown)


bundlewrap/metadata.py
```diff
@@ -24,6 +25,39 @@ class DoNotRunAgain(Exception):
...
+    """
+    if isinstance(obj, METADATA_TYPES):
+        return obj
+    elif isinstance(obj, dict):
+        if isinstance(obj, ATOMIC_TYPES[dict]):
+            new_obj = atomic({})
+        else:
+            new_obj = {}
+        for key, value in obj.items():
+            new_key = copy(key)
```
vain, line 40: Any particular reason why this no longer checks if `key` is in `METADATA_TYPES`, too? Users can now sneak in unsupported types.

trehn: They can't because `validate_metadata()` now exists independently and is called from `._set_layer()`.

PR #505, opened 2020-05-19
[RFC] Faults get identifiers to be comparable

vain:
I am not happy with this. It is better than before, but it is very easy to accidentally create collisions, thus making unequal Faults “equal”. It’s just not a fool proof interface.

But how else would you make Faults comparable?

### DIFF
--- a/bundlewrap/secrets.py
+++ b/bundlewrap/secrets.py
@@ -251,6 +251,7 @@ class SecretProxy:
 
     def decrypt(self, cryptotext, key=None):
         return Fault(
+            'bw secrets decrypt {} {}'.format(cryptotext, key),
             self._decrypt,
             cryptotext=cryptotext,
             key=key,
@@ -258,6 +259,7 @@ class SecretProxy:
 
     def decrypt_file(self, source_path, key=None):
         return Fault(
+            'bw secrets decrypt_file {} {}'.format(source_path, key),
             self._decrypt_file,
             source_path=source_path,
             key=key,
@@ -265,6 +267,7 @@ class SecretProxy:
 
     def decrypt_file_as_base64(self, source_path, key=None):
         return Fault(
+            'bw secrets decrypt_file_as_base64 {} {}'.format(source_path, key),
             self._decrypt_file_as_base64,
             source_path=source_path,
             key=key,
@@ -319,6 +322,13 @@ class SecretProxy:
         self._call_log.setdefault(identifier, 0)
         self._call_log[identifier] += 1
         return Fault(
+            'bw secrets human_password {} {} {} {} {}'.format(
+                identifier,
+                digits,
+                key,
+                per_word,
+                words,
+            ),
             self._generate_human_password,
             identifier=identifier,
             digits=digits,
@@ -331,6 +341,12 @@ class SecretProxy:
         self._call_log.setdefault(identifier, 0)
         self._call_log[identifier] += 1
         return Fault(
+            'bw secrets password {} {} {} {}'.format(
+                identifier,
+                key,
+                length,
+                symbols,
+            ),
             self._generate_password,
             identifier=identifier,
             key=key,
@@ -340,6 +356,11 @@ class SecretProxy:
 
     def random_bytes_as_base64_for(self, identifier, key='generate', length=32):
         return Fault(
+            'bw secrets random_bytes_as_base64 {} {} {}'.format(
+                identifier,
+                key,
+                length,
+            ),
             self._generate_random_bytes_as_base64,
             identifier=identifier,
             key=key,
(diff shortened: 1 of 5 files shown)


bundlewrap/secrets.py
```diff
@@ -319,6 +322,13 @@ def human_password_for(
         self._call_log.setdefault(identifier, 0)
         self._call_log[identifier] += 1
         return Fault(
+            'bw secrets human_password {} {} {} {} {}'.format(
```
vain, line 325: A bit ugly. One might think about automatically adding kwargs to the fault_identifier. I haven’t done that because, well, maybe one of those kwargs is a Fault? Who knows.

vain:
We currently have this in our repo:

```python
def wrap(something):
    def callback():
        return something
    return Fault(callback)
```

The point is to ensure the Fault API (e.g., `format_into`) is available.

This would have to be rewritten as:
```python
def wrap(something):
    def callback():
        return something

    if isinstance(something, Fault):
        fi = Fault.id_list
    else:
        fi = [something]

    return Fault(fi + ['wrap'], callback)
```
```

**Comment by vain** [comment]
We also have this:
```python
def dict_as_json(some_dict):
    def dump_the_dict():
        return dumps(
            some_dict,
            cls=MetadataJSONEncoder,
            indent=4,
            sort_keys=True,
        )
    return Fault(dump_the_dict)
```
I’m stumped, for the moment.

bundlewrap/secrets.py
```diff
@@ -340,6 +356,11 @@ def password_for(self, identifier, key='generate', length=32, symbols=False):
 
     def random_bytes_as_base64_for(self, identifier, key='generate', length=32):
```
trehn, line 357: In theory, this is unsafe for key names with spaces in them.
```python
random_bytes_as_base64_for("foo", key="bar baz") == random_bytes_as_base64_for("foo bar", key="baz")
```

trehn: should probably just hash both strings?

bundlewrap/secrets.py
trehn: Don't get it. `digits` might be a fault now, so what?

trehn:
>dump_the_dict

Lucky for us, this is only used in items.py, where comparing Faults is no longer relevant. So we can just pass a static identifier and be done with it. Maybe call it `dump_the_dict_only_use_in_itemspy` :D

trehn:
>it is very easy to accidentally create collisions

Hm. How? Most people will use the builtin functions and should be safe.

vain:
> Hm. How? Most people will use the builtin functions and should be safe.

Yeah. Let’s hope so.

> Don't get it. `digits` might be a fault now, so what? […] should probably just hash both strings?

Alright, `Fault` now always adds all kwargs (hashed) to `id_list`. Means we no longer have to construct those ugly strings.

However, since dicts are not hashable in Python, it means you can no longer pass a dict when constructing a new Fault. Is that a problem? I hope not.

bundlewrap/utils/__init__.py
```diff
@@ -96,21 +105,20 @@ def __add__(self, other):
         if isinstance(other, Fault):
             def callback():
                 return self.value + other.value
-            return Fault(callback)
+            return Fault(self.id_list + other.id_list, callback)
         else:
             def callback():
                 return self.value + other
-            return Fault(callback)
+            return Fault(self.id_list + ['raw {}'.format(other)], callback)
```
trehn, line 112:
```suggestion
            return Fault(self.id_list + ['raw {}'.format(repr(other))], callback)
```

Otherwise you cannot tell the difference between `other` being  `1` or `"1"`.

trehn: not a fan of the `repr()` here, but it seems the simplest way to catch most issues

vain: Applied.

bundlewrap/utils/__init__.py
```diff
@@ -74,7 +74,16 @@ class Fault:
...
-    def __init__(self, callback, **kwargs):
+    def __init__(self, fault_identifier, callback, **kwargs):
+        if isinstance(fault_identifier, list):
+            self.id_list = fault_identifier
+        else:
+            self.id_list = [fault_identifier]
+
+        for key, value in sorted(kwargs.items()):
+            self.id_list.append(hash(key))
+            self.id_list.append(hash(value))
```
trehn, line 85: `hash()` return values only remain constant within the same process. So this would break if we ever pickled a Fault and imported it into another process. But that is probably not going to happen.

trehn:
LGTM, can you update the docs (removing the warning for metadata.py) and merge?

PR #506, opened 2020-05-19
add `bw metadata --colors` and `--hide-defaults`

### LINKED ISSUE #496: bw metadata: CLI option to hide defaults

trehn:
closes #496 

Should colors be on by default?

### DIFF
--- a/bundlewrap/utils/text.py
+++ b/bundlewrap/utils/text.py
@@ -57,11 +57,6 @@ def green(text):
 
 
 @ansi_wrapper
-def grey(text):
-    return "\033[90m{}\033[0m".format(text)
-
-
-@ansi_wrapper
 def red(text):
     return "\033[31m{}\033[0m".format(text)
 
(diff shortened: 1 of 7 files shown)


bundlewrap/utils/text.py
```diff
@@ -56,6 +56,11 @@ def green(text):
     return "\033[32m{}\033[0m".format(text)
 
 
+@ansi_wrapper
+def grey(text):
+    return "\033[90m{}\033[0m".format(text)
```
vain, line 61: That’s problematic as it doesn’t produce meaningful output when using a bright terminal color scheme. Better go with blue.

bright terminal

vain:
Weeeeee! I’m very much looking forward to this.

> Should colors be on by default?

I’d do the usual “colorize if stdout is a tty or forced”.

vain:
LGTM … except for those failing tests. %)

PR #521, opened 2020-06-30
git_deploy: Do not use shallow clones for HTTP

vain:
HTTP does not support that.

Port of https://github.com/bundlewrap/plugins/pull/11.


vain:
```
$ git clone --bare --depth 1 --no-single-branch https://uninformativ.de/git/xlonitor.git
Cloning into bare repository 'xlonitor.git'...
warning: redirecting to https://www.uninformativ.de/git/xlonitor.git/
fatal: dumb http transport does not support shallow capabilities
```

PR #527, opened 2020-07-15
Metadata: Only report clashes if values are not equal

vain:
Makes it possible to return identicals 'defaults' from different bundles
or do more complicated stuff like this:

    @metadata_reactor
    def generic_default(metadata):
        # A generic default that works for most system. (Yes, this
        # particular case could be solved without a reactor. Think of a more
        # complicated example, where this function depends on other factors.)
        def_iface = list(metadata.get('interfaces').keys())[0]

        return {
            'internal_interface': metadata.get('internal_interface', def_iface),
        }

    @metadata_reactor
    def corner_case(metadata):
        # This is only present on some special systems.
        special_iface = 'eth2'

        if metadata.get(f'interfaces/{special_iface}', None) is not None:
            return {
                'internal_interface': special_iface,
            }
        else:
            return {}

### DIFF
--- a/bundlewrap/metadata.py
+++ b/bundlewrap/metadata.py
@@ -108,21 +108,21 @@ def check_for_metadata_conflicts_between_defaults_and_reactors(node):
         for path in map_dict_keys(d):
             value = value_at_key_path(d, path)
             if isinstance(value, dict):
-                yield path, TYPE_DICT
+                yield path, value, TYPE_DICT
             elif isinstance(value, set):
-                yield path, TYPE_SET
+                yield path, value, TYPE_SET
             else:
-                yield path, TYPE_OTHER
+                yield path, value, TYPE_OTHER
 
     for prefix in ("metadata_defaults:", "metadata_reactor:"):
         paths = {}
         for identifier, layer in node._metadata_stack._layers.items():
             if identifier.startswith(prefix):
-                for path, current_type in paths_with_types(layer):
+                for path, value, current_type in paths_with_types(layer):
                     try:
-                        prev_type, prev_identifier = paths[path]
+                        prev_type, prev_identifier, prev_value = paths[path]
                     except KeyError:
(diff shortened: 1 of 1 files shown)


trehn:
Blame will still show only one of the possible sources, right?

vain:
Blame shows both:

```
@metadata_reactor
def r1(metadata):
    return {
        'clash': True,
    }


@metadata_reactor
def r2(metadata):
    return {
        'clash': True,
    }
```
```
$ bw metadata --blame node-1 -k clash
╭───────┬────────────────────────────────────────────────╮
│ path  │ source                                         │
├───────┼────────────────────────────────────────────────┤
│ clash │ metadata_reactor:aa.r1, metadata_reactor:aa.r2 │
╰───────┴────────────────────────────────────────────────╯
```
A simple `bw metadata node-1` colors that key green in this scenario and also green if I turn `r1` into `defaults`. Both correct, IMHO.

vain:
Just for the record, the code above is very wrong:

```
@metadata_reactor
def generic_default(metadata):
    # A generic default that works for most system. (Yes, this
    # particular case could be solved without a reactor. Think of a more
    # complicated example, where this function depends on other factors.)
    def_iface = list(metadata.get('interfaces').keys())[0]

    return {
        'internal_interface': metadata.get('internal_interface', def_iface),
    }
```

In particular, this pattern is the problem:
```
return {'foo': metadata.get('foo', something_that_depends_on_metadata)}
```
* Let’s assume that on the first run `foo` does not yet exist in metadata. So, we create it and set it to our default.
* During all follwing runs, we will *always return that same value*, even if the other metadata *which we wanted to react to* has changed.

What to do instead? I’m not sure, yet. The *intention* is: Create some metadata *default*, but you can’t put it into the real `defaults`, because it depends on other metadata. It is supposed to be “overwriteable” by another reactor.

PR #529, opened 2020-07-16
rewrite and optimize metadata generation

trehn:


trehn:
almost 50% of runtime is spent in `Metastack.get()`, we should flatten defaults and node/group metadata into a single layer each before running reactors

bundlewrap/utils/metastack.py
```diff
@@ -42,18 +44,21 @@ def get(self, path, default=NO_DEFAULT):
...
+            # prefer cached partitions if available
+            partition = self._cached_partitions.get(part_index, partition)
+            for layer in partition.values():
+                try:
+                    value = value_at_key_path(layer, path)
+                except KeyError:
+                    pass
                 else:
-                    result = merge_dict(result, {'data': value})
+                    if undef:
```
vain, line 56: Another idea for an optimization: If we see an atomic value here, we exit early, because that's the final result.

trehn: Haha, was working on that already. But it's not that simple because we'd need to reverse the order we look at layers. The current code is working bottom-up, but for an early return we need to go top-down.

vain: Yeah, too bad, the previous version before the force-push implied that we already went top to bottom …

PR #532, opened 2020-07-22
Metastack: Only return early if this is the first value we see

vain:
The old code returned UNMERGEABLE values if they showed up in lower
layers. See failed tests.

### DIFF
--- a/bundlewrap/utils/metastack.py
+++ b/bundlewrap/utils/metastack.py
@@ -56,10 +56,11 @@ class Metastack:
                 except KeyError:
                     pass
                 else:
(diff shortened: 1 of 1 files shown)


vain:
https://travis-ci.org/github/bundlewrap/bundlewrap/jobs/710651711

vain:
In my world view, it should be a type error if we see an UNMERGEABLE value only in a lower layer.

We nuked `MetastackTypeConflict` at some point. Do you remember why? :)

vain:
Builds still failing for some Python versions, looks related to this: https://github.com/bundlewrap/bundlewrap/pull/533

PR #533, opened 2020-07-22
Avoid Python segfaults

vain:
For reasons beyond my understanding, this caused segfaults on our build
agent with Ubuntu's Python 3.8.0. Doesn't happen on my local machine
with Arch Linux's Python 3.8.3.


vain:
Travis just spat out this with Python 3.7.7:

```
>           for layer in reversed(partition.values()):
E           TypeError: 'dict_values' object is not reversible
```

PR #534, opened 2020-07-28
Make message about illegal metadata more helpful

vain:
If you put floats in there, it just said "illegal value: 20.04" and
nobody understood why.


vain:
I was thinking about listing all legal types there, would that help? Right now, people must check the docs.

PR #541, opened 2020-08-24
Fault: Hash kwargs recursively

vain:
Mutable objects are not hashable in Python. We'd like to pass those as
arguments to Faults, though. For example:

    def whatever(data):
        return json.dumps(data)

    f = Fault('foo', whatever, data={'what': 'ever', 'some': other_fault})

Mutable objects are not hashable, because, well, they're mutable. You
cannot determine a hash, because that object might change later
in-place.

Strictly speaking, this is a problem for us, too:

    def my_function(data):
        return f'{repr(data)}'

    dirty_dict = {
        'foo': 0,
    }
    f = Fault('whatever', my_function, data=dirty_dict)

    dirty_dict['foo'] = 1
    g = Fault('whatever', my_function, data=dirty_dict)

    print(f'Will print False: {f == g}')
    print(f'Will print same string twice, which is wrong: {str(f)}, {str(g)}')

Let's discuss whether we should do deepcopy kwargs or not.

### DIFF
--- a/bundlewrap/utils/__init__.py
+++ b/bundlewrap/utils/__init__.py
@@ -113,7 +113,7 @@ class Fault:
 
         for key, value in sorted(kwargs.items()):
             self.id_list.append(hash(key))
-            self.id_list.append(value)
+            self.id_list.append(_recursive_hash(value))
 
         self._available = None
         self._exc = None
@@ -149,7 +149,7 @@ class Fault:
             return self.id_list == other.id_list
 
     def __hash__(self):
-        return _recursive_hash(self.id_list)
+        return hash(tuple(self.id_list))
 
     def __len__(self):
         return len(self.value)
--- a/bundlewrap/utils/__init__.py
+++ b/bundlewrap/utils/__init__.py
@@ -113,7 +113,7 @@ class Fault:
 
         for key, value in sorted(kwargs.items()):
             self.id_list.append(hash(key))
-            self.id_list.append(hash(value))
+            self.id_list.append(_recursive_hash(value))
 
         self._available = None
         self._exc = None
@@ -202,6 +202,25 @@ for method_name in (
     'zfill',
 ):
     setattr(Fault, method_name, _make_method_callback(method_name))
+
+
+def _recursive_hash(obj):
+    hashes = []
+    if isinstance(obj, list):
+        for i in obj:
+            hashes.append(_recursive_hash(i))
+        return hash(tuple(hashes))
+    elif isinstance(obj, set):
+        for i in sorted(obj):
+            hashes.append(_recursive_hash(i))
+        return hash(tuple(hashes))
+    elif isinstance(obj, dict):
+        for k, v in sorted(obj.items()):
+            hashes.append(hash(k))
+            hashes.append(_recursive_hash(v))
+        return hash(tuple(hashes))
+    else:
+        return hash(obj)
 
 
 def get_file_contents(path):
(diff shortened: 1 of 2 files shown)


vain:
Possible arguments against deepcopy:

* When passing a closure as callback, we have no control anyway. Users can reference a mutable object there and change it later on, thus breaking things.
* In practice, it's probably good enough without deepcopy. (Weak argument.)

trehn:
Using deepcopy here is also inherently unpythonic. This is Python and for better or worse, people can expect to change things after they stored them somewhere else (in a Fault). We should not create things that change this fundamental property of Python code.

I'd much rather have a slighty more complex definition of Faults being equal: when they were passed the same callback and parameters. The key word here being "passed". The moment of Fault creation counts.

It is easy to be led astray by our natural preference for pure functions. However, there is no guarantee anyway that our callback is a pure function. Therefore, Fault equality cannot be tied to the value the Fault eventually resolves to. It can only be determined by the information we get upon creating the Fault.

And where would be put deepcopy anyway? We're already assembling `id_list` in `__init__` anyway, so there is no opportunity for mutable objects to change.

vain:
(Maybe I'm misunderstanding, as I often do. :))

> We're already assembling id_list in __init__, so there is no opportunity for mutable objects to change after being hashed.

Yes, there is, and that's kind of my concern.

You pass an object to a Fault constructor, which builds `id_list`. That `id_list` will never change again -- but the objects passed to the constructor can change. This *can* lead to inconsistencies. The second code example above demonstrates that.

I agree that it's rare and that there are many ways to screw up. Not saying we *must* deepcopy (I'd do that in the constructor, at the same time when we're building `id_list`), just trying to spot potential pitfalls. If documented properly (uh, well, there's little doc on Faults anyway), I'd be happy with the current code of this PR.

trehn:
>so there is no opportunity for mutable objects to change after being hashed.

That was poorly worded. Obviously, yes, they can change. But that won't affect the hash and thus won't violate my stipulation that Fault equality is determined at the time of `__init__` based on the information available at that time.

Your answer gave me another idea though: Why don't we put the actual objects in `id_list` instead of their hashes? Obviously that makes checks for equality a rather expensive operation (though we could be smart by using an iterator over the list...). But then Faults would compare equal if the *current* state of their parameters is equal.

trehn:
What do we need `__hash__` for?

vain:
> What do we need __hash__ for?

So that Faults can be used in sets. (I wondered that too for a minute. Goldfischgedächtnis™.)

trehn:
Ugh, I'm sorry, but we need to revert c67f5f3.

From https://docs.python.org/3/glossary.html#term-hashable:

> An object is hashable if it has a hash value which **never changes during its lifetime** (it needs a `__hash__()` method), and can be compared to other objects (it needs an `__eq__()` method). Hashable objects which compare equal must have the same hash value.

I think building the hash once based on the initial state of the arguments is the only valid solution.

vain:
Yeah.

None of this is satisfying. As you said, a Fault’s callback could do virtually anything, including returning a random value, and we have no way of catching that. It all depends on whether the user screws up or not. Can lead to subtle bugs that take ages to track down.

I’d still love to see Faults being removed some day. Go back to magic strings that are only ever resolved in templates.

Should we add more documentation on this in another PR? Right now, none of this is documented. Faults are only explained as “`repo.vault.password_for` returns them”. For the purpose of core bw, I think this is not that bad, actually. People shouldn’t use them directly. :) Maybe we should just add internal documentation at our company.

PR #545, opened 2020-09-06
fix typo in docs

veecue:
The attribute of ItemStatus is called keys_to_fix instead of keys. This is fixed in the example item comment.


trehn:
Thanks!

Issue #542, opened 2020-08-31
Deploy to non-root environment


veecue:
I would like to deploy accounts on shared servers where I don't have root access.
| Problem | Workaround | Suggested fix|
|--|--|--|
|Locks do not work|overwrite `bundlewrap.lock.LOCK_BASE` in an `apply_start` hook|Set `bundlewrap.lock.LOCK_BASE` to "~/.bundlewrap" by default or allow configuration|
|No filepath relative to home|Hardcode homedir in metadata|Allow filepaths starting with `~/`|
|Files are owned by root by default|Hardcode username and group in metadata and set for every file instance|Make default owner/group `None` (don't do chown at all)|

And of course I had to change `'cmd_wrapper_outer'` to not use sudo (which I would do anyway because I ssh using root@ and often don't have sudo installed).

With those fixes, the described usecase should be supported quite well. What do you think?

veecue:
Quoting `bw zen`:

> BundleWrap is the single point of truth.

This is rather contradicting with the idea of relative filepaths and dynamic users. I use `action` items that rely on the current dir which is bad enough. Probably the workarounds for those two problems are the more elegant solution.

PR #546, opened 2020-09-08
allow lock dir to be configured per node

veecue:
Add the `lock_dir` property that dictates where locks are stored on a
specific node. The peviously hardcoded `/var/lib/bundlewrap` is used as
a default. With this change, bundlewrap can be used on nodes, where
/var/lib is not acessable for some reason (e.g. mounted readonly).

This change also allows for usage on nodes without root access (#542)

I tested this manually using bw lock and bw apply. If you want an automated test, let me know!

Also I'm unsure if this change qualifies as IP.


trehn:
Thank you!

Issue #524, opened 2020-07-06
pkg_pip fails if there is only `pip3`


fkusei:
Maybe bw could check if `pip3` is available, if `pip` fails?
Or maybe the default should be to use `pip3`?

Semi-related to #396

PR #547, opened 2020-09-12
items/pkg_pip: allow configuring pip_command

Kunsi:
This allows to customize the call to `pip` in `pkg_pip:` items to fix the issue discussed in #524.

I thought about changing the default to pip3, but that may break existing repos, so i've decided against that.


trehn:
Thanks!

Issue #540, opened 2020-08-24
Feature Request: List of things which should be deployed as early as possible during apply


fkusei:
We get a lot of 'sudo: cannot resolve hostname' emails when doing an initial apply of a node. Some time during an apply, a `/etc/hosts` file gets deployed, which then allows us to resolve the host name of this machine.

Ideal would be to deploy this file as early as possible during an apply, to avoid getting most of those mails.

veecue:
What I do for this is create a tag called "init" (with proper dependencies within the tag) and then deploy a new node using `bw apply node-1 -o tag:init` first

fkusei:
Yeah, using a tag would work - but we're also talking changing node hostnames and stuff. Would like to avoid having to do two applies :/

veecue:
Another possibility I can think of would be adding a global item attribute called `stage` or `priority` or something like that and then have all items of a stage depend on all ones of the previous stage. This could either be numbers (0 would be the default, negative some init stuff and positive some check/test/finalize stuff). Maybe also using texts (`pre-init`, `init`, `finalize`, `post-finalize`), would be less flexible but easier to use. Having a `needs` to a higher stage would have to raise an error in `bw test` then.

trehn:
resolved in #559

PR #550, opened 2020-09-18
items: Add stages

veecue:
Item's can have a "stage" property. All items on one specific stage
depdend on all items of all previous stages. Stages are integers with
the default stage being zero.

Example usecases are important initialization steps (that would have a
negative stage) and integration testing steps (that would have a
positive stage).

Fixes: #540

TODOs:
 - [x] Check with @fkusei if this suits their needs
 - [ ] Check with @trehn if the general concept is ok for bundlewrap
 - [ ] Add documentation
 - [ ] Add tests
 - [ ] Handle in `bw plot node` somehow (maybe color codes, boxes like bundles or at least allow hiding all those depndencies
 - [ ] (optional) Maybe create the following dependencies to generate nicer plots: `<items of stage 0> --> stage0post --> stage1pre --> <items of stage 1> --> stage1post --> stage2pre --> <items of stage 2>` instead of `<items of stage 0> --> stage0 --> <items of stage 1> --> <stage 1> --> <items of stage 2> --> <stage 2>`
 - [ ] fix skipping dependencies

### DIFF
--- a/bundlewrap/deps.py
+++ b/bundlewrap/deps.py
@@ -20,6 +20,7 @@ class DummyItem:
         self.tags = []
         self.triggered_by = []
         self.triggers = []
+        self.stage = 0
         self._deps = []
         self._precedes_items = []
 
@@ -89,6 +90,24 @@ class TypeItem(DummyItem):
     def id(self):
         return "{}:".format(self.item_type)
 
+class StageItem(DummyItem):
+    """
+    This item depends on all items with the given stage
+    and on the previous stage.
+    """
+    ITEM_TYPE_NAME = 'stage'
+
+    def __init__(self, stage):
+        super(StageItem, self).__init__()
+        self.stage = stage
+        self.prev = None
+
+    def __repr__(self):
+        return "<StageItem: {}>".format(self.stage)
+
+    @property
+    def id(self):
+        return "stage:{}".format(self.stage)
 
 def find_item(item_id, items):
     """
@@ -368,6 +387,29 @@ def _inject_tag_items(items):
     items.update({item.id: item for item in tag_items.values()})
     return items
 
+def _inject_stage_items(items):
+    """
+    Takes a list of items and adds stage items depending on all
+    items of a stage and the previous stage
+    """
+    stage_items = {}
+    for item in items.values():
+        if item.stage not in stage_items:
+            stage_items[item.stage] = StageItem(item.stage)
+    if len(stage_items) == 1:
+        # Don't need to add depnedencies for only one stage
+        return items
+    prev_id = None
+    for stage_item in sorted(stage_items.values(), key=lambda stage_item: stage_item.stage):
+        if prev_id is not None:
+            stage_item.prev = prev_id
+        prev_id = stage_item.id
+    for item in items.values():
+        stage_items[item.stage]._deps.append(item.id)
+        if stage_items[item.stage].prev is not None:
+            item._deps.append(stage_items[item.stage].prev)
+    items.update({item.id: item for item in stage_items.values()})
+    return items
 
 def _inject_type_items(items):
     """
@@ -602,6 +644,7 @@ def prepare_dependencies(items, node_os, node_os_version):
 
     items = _inject_bundle_items(items)
     items = _inject_tag_items(items)
+    items = _inject_stage_items(items)
     items = _inject_type_items(items)
     items = _inject_canned_actions(items)
     items = _inject_reverse_triggers(items)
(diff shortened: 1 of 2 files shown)


fkusei:
I think that would be a way to implement what I was looking for. However i don't feel competent enough to judge your code, i'll pass :)

(PS: It's 'her' or 'their', not 'his')

veecue:
@fkusei Well, sorry for assuming your gender. Probably that happened because I'm used to the german male-neutral.

trehn:
I'm very conflicted. Here are some thoughts:

The functional overlap with tags is not as large as I thought. Even if we had a default tag on every item, we couldn't replicate this functionality. 

Using numerical stages invites undocumented, implicit dependencies while BundleWrap has always strived to be very explicit with these things. Docs should very clearly recommend explicit dependencies and discourage over-use of stages.

Why does this item belong in this stage? Can I move it somewhere else? What other items are in that stage, anyway? Can we somehow make this more self-documenting?

The problem laid out in #540 could also be solved by a pre-apply hook. Are there really enough use cases that require items instead of hooks?

Listing any StageItem in a regular `needs` or `needed_by` should probably be an error. `triggers` however is more interesting. We probably should support all kinds of DummyItems as trigger targets.

`precedes` is always a headache and I don't even want to think about it right now.

`Item.covered_by_autoskip_selector` and `Item.covered_by_autoonly_selector` need to be updated to understand stages.

Stage transitions should probably be visible during `bw apply`, especially for the cases where we're not just dealing with the typical 3-stage pre/main/post situation, but rather some convoluted deployment process.

Is stage really an item-level property or will users want to assign entire bundles to stages?

@vain Care to weigh in?

veecue:
I'm also totally unconvinced by my current implementation. Some more thoughts/replies:

> The problem laid out in #540 could also be solved by a pre-apply hook. Are there really enough use cases that require items instead of hooks?

I guess there are no use cases that really _require_ items, but as soon as you need files or even templates, it just makes more sense. Let me lay out how I would split my process into stages as an example
- -3: Mask a failing but non applicable systemd unit because I use `systemctl is-system-running` for sanity checks a few times
- -2: Set up mirrors, networking and everything needed to install packages
- -1: (optionally) do a full system update: This is particularly interesting for stages since it conflicts with all other items automatically, especially the reboot. Also doing an update might be required before installing packages to make sure they are available on the server.
- 0: Do all normal deployments etc.

The usecase for positive numbers I can think of would be some kind of UAT test suite, but I guess hooks make more sense there.

> Why does this item belong in this stage? Can I move it somewhere else? What other items are in that stage, anyway? Can we somehow make this more self-documenting?

One idea I had to solve that problem would be to force the user to define named stages and their order. However there is currently no repo-wide config file I could imagine for this purpose.

> Listing any StageItem in a regular `needs` or `needed_by` should probably be an error. `triggers` however is more interesting. We probably should support all kinds of DummyItems as trigger targets.
> 
> `precedes` is always a headache and I don't even want to think about it right now.
> 
> `Item.covered_by_autoskip_selector` and `Item.covered_by_autoonly_selector` need to be updated to understand stages.
> 
> Stage transitions should probably be visible during `bw apply`, especially for the cases where we're not just dealing with the typical 3-stage pre/main/post situation, but rather some convoluted deployment process.
> 
> Is stage really an item-level property or will users want to assign entire bundles to stages?

All those concerns come from the current implementation. One addional problem I found is that as any item is skipped in interactive mode, all following stages are just skipped.

Maybe a version that is easier to understand, more self-documenting and cleaner to implement would be to assign bundles to stages on a node/group-level. Then the apply process would have to be chopped into stages that are totally independent from one another instead of realizing the dependencies on a per-item basis. Although I decided against this, it might have way less rabbitholes in the end.

vain:
Random thoughts:

Item dependencies are a massive nightmare in larger repos. In general. They’re quite in-transparent, because anything can depend on anything. That’s powerful, but hard to master. Gets worse when people use dependencies on `pkg_apt:` or similar. Plotting dependencies is useless.

Stages (like targets in systemd) look promising. Feels like this could bring order to chaos. I like the idea. **But**: This would probably require people to reorganize their repos, like, a lot. You have to re-think everything in order to make good use of stages. Stages are not a quick fix for isolated problems. You must create very clear definitions of each stage. Just look at systemd, this can get complicated really fast. And then you basically have to assign a stage to each item manually.

Stages must have names, not numbers. We would probably end up copying systemd’s model of `default.target`.

Why can’t stages be implemented by using tags?

At the moment, I’m quite fond of stages (be it tags or not). It’s a lot of work (not for bw core, but the user’s repos), but it feels like the right thing to do.

trehn:
@veecue I have drafted an alternative concept for stages in #553 and will probably start hacking on that soon. Would that concept work for you?

trehn:
Thanks for the effort, but I'm closing this in favor of the upcoming changes as detailed in #553

PR #552, opened 2020-09-25
hooks: call apply_start before parsing items

veecue:
When setting the progress total by using count_items() before calling
the `apply_start` hook, all bundles and items will be parsed. That might
be undesirable if a hook should make changes to the local state that
would reflect in an item's state. Therefore we should set the progress
after calling the apply_start hook.

This change is backwards compatible because if a hook actually acesses
some of the bundles/items in the repo, they will be dynamically parsed
on demand thanks to the use of @cached_property.


trehn:
Thanks!

PR #558, opened 2020-10-23
cmdline/{apply,run}: sort node list in stats table

Kunsi:
This will sort the node list in the stats table. I think this is much more useful than the current "random" order (which seems to be the order in which the nodes finished processing?).

### DIFF
--- a/bundlewrap/cmdline/run.py
+++ b/bundlewrap/cmdline/run.py
@@ -73,6 +73,7 @@ def stats_summary(results, include_stdout, include_stderr):
     if include_stderr:
         rows[0].append(bold(_("stderr")))
 
+    node_rows = []
     for node_name, result in sorted(results.items()):
         row = [node_name]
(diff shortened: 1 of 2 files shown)


bundlewrap/cmdline/run.py
```diff
@@ -73,6 +73,7 @@ def stats_summary(results, include_stdout, include_stderr):
     if include_stderr:
         rows[0].append(bold(_("stderr")))
 
+    node_rows = []
     for node_name, result in sorted(results.items()):
```
trehn, line 77: run results are already sorted here

trehn:
Thank you, but I found an easier way to sort the apply results: https://github.com/bundlewrap/bundlewrap/commit/cbd449b05f6b425e1b3558d8b945c6cbd4fe63ed

Kunsi:
Looks better than mine, yeah :(

PR #561, opened 2020-11-08
utils/Fault: support using {fault} in format_into()

Kunsi:
This allows people to use the fault in multiple positions in the format string.

I needed this to be able to generate a unique, but rotates-with-secret-rotation identifier, which will be used as name for dkim signing keys.

This is the code i've written: https://git.kunsmann.eu/kunsi/bundlewrap/commit/f6d6fd79587406852e6d992321befaf2fcf2d2f7


trehn:
Sorry for the late response -_-

You can do this already with `{0}`.

PR #563, opened 2020-11-16
Revert "k8s: Ignore 'apiVersion' when comparing state"

vain:
This reverts commit a05b24b5edbbe5a10ed03c1005edb324f8bdc631.

This was wrong. You must have matching apiVersions. You just cannot
patch an existing object when apiVersions differ, because object schemas
(can) change between apiVersions.

### DIFF
--- a/bundlewrap/items/kubernetes.py
+++ b/bundlewrap/items/kubernetes.py
@@ -54,7 +54,7 @@ class KubernetesItem(Item, metaclass=ABCMeta):
             return None
         else:
             return {'manifest': json.dumps(
-                self.nuke_changing_attributes(json.loads(self.manifest)),
+                self.nuke_k8s_status(json.loads(self.manifest)),
                 indent=4, sort_keys=True,
             )}
 
@@ -148,12 +148,9 @@ class KubernetesItem(Item, metaclass=ABCMeta):
     def namespace(self):
         return self.name.split("/", 1)[0] or None
 
-    def nuke_changing_attributes(self, manifest):
-        # These attributes cannot be managed by us. The cluster decides
-        # what's going on. We can't fix it.
-        for i in ('apiVersion', 'status'):
-            if i in manifest:
-                del manifest[i]
+    def nuke_k8s_status(self, manifest):
+        if 'status' in manifest:
+            del manifest['status']
         return manifest
 
     def patch_attributes(self, attributes):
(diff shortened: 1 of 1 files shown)


vain:
If you ask me, this should be a hard error. It’s not a fixable attribute, it’s a mismatch. Should I throw an exception in `sdict()` if `apiVersion` doesn’t match?

trehn:
No, sdict should still be able to retrieve the remote state without an error. `fix()` has to decide it can't do anything. This is an unprecedented case and there are no mechanisms in place to elegantly deal with it.

vain:
Alright, then this PR could be merged as is.

Without the PR, you have no idea what’s going on:

```diff
? k8s.dev  ╭─ k8s_crd:orders.acme.cert-manager.io
? k8s.dev  │
? k8s.dev  │  manifest
? k8s.dev  │  @@ -19,8 +19,40 @@
? k8s.dev  │           "name": "orders.acme.cert-manager.io"
? k8s.dev  │       },
? k8s.dev  │       "spec": {
? k8s.dev  │  +        "additionalPrinterColumns": [
? k8s.dev  │  +            {
? k8s.dev  │  +                "JSONPath": ".status.state",
? k8s.dev  │  +                "name": "State",
? k8s.dev  │  +                "type": "string"
? k8s.dev  │  +            },
```

With the PR, you at least get a hint:

```diff
? k8s.dev  ╭─ k8s_crd:orders.acme.cert-manager.io
? k8s.dev  │
? k8s.dev  │  manifest
? k8s.dev  │  @@ -1,5 +1,5 @@
? k8s.dev  │   {
? k8s.dev  │  -    "apiVersion": "apiextensions.k8s.io/v1",
? k8s.dev  │  +    "apiVersion": "apiextensions.k8s.io/v1beta1",
? k8s.dev  │       "kind": "CustomResourceDefinition",
? k8s.dev  │       "metadata": {
? k8s.dev  │           "annotations": {
? k8s.dev  │  @@ -20,8 +20,40 @@
? k8s.dev  │           "name": "orders.acme.cert-manager.io"
? k8s.dev  │       },
? k8s.dev  │       "spec": {
? k8s.dev  │  +        "additionalPrinterColumns": [
? k8s.dev  │  +            {
? k8s.dev  │  +                "JSONPath": ".status.state",
? k8s.dev  │  +                "name": "State",
? k8s.dev  │  +                "type": "string"
? k8s.dev  │  +            },
```

The thing is, the diff can be massive. It’s hard to spot those two lines that tell you about the version mismatch.

Other than that, affected items will always be “failed”, which is good.

PR #574, opened 2021-01-04
Optimize metadata generation through @metadata_reactor.provides()

trehn:


bundlewrap/metagen.py
```diff
@@ -19,6 +19,180 @@
...
+def reactors_for_paths(available_reactors, required_paths):
+    """
+    Returns only those available_reactors that might affect the
+    required_paths.
+    """
+    for name, reactor in available_reactors:
+        provides = getattr(reactor, '_provides', tuple())
+        if provides:
+            for path in provides:
+                if required_paths.covers(path):
```
vain, line 71: Shouldn’t this be the other way around? The path from the decorator should cover one of the required paths?

vain:
Still seeing a couple of crashes in our big repo. I’m in the process of debugging.

bundlewrap/metagen.py
trehn: It's actually both ways:

```
metadata.get('foo') -> required_paths == {'foo'}
@metadata_reactor.provides('foo/bar') -> foo covers foo/bar, so this reactor is relevant
```

```
metadata.get('foo/bar') -> required_paths == {'foo/bar'}
@metadata_reactor.provides('foo') -> reactor might provide foo/bar, so it's relevant
```

But you're right, the latter case is not handled properly.

trehn: pushed a quick fix

vain: Aha! Crashes gone. Metadata hash remained the same. 🎉

PR #579, opened 2021-01-13
bundlewrap.utils.Fault: add .as_htpasswd_entry()

### LINKED ISSUE #578: [Feature Request] vault.password_for() should be able to return a htaccess-compatible, encrypted password
It would be nice if one could generate htaccess-compatible encrypted password lines without handling Faults themselves.

We're currently using something like this:
```
from passlib.hash import apr_md5_crypt
from hashlib import sha512
from bundlewrap.utils import Fault

pasword = Fault(
    'some_identifier',
    lambda pw: apr_md5_crypt.encrypt(
        str(pw),
        salt=sha512(node.name.encode('utf-8')).hexdigest()[:8]
    ),
    pw=pw
)
```

fkusei:
This will allow users to easily generate htpasswd-compatible `username:password` lines. Similar functionality is already present in trehn/bundlewrap-teamvault for teamvault entries.

This pull request will fix #578

### DIFF
--- a/bundlewrap/utils/__init__.py
+++ b/bundlewrap/utils/__init__.py
@@ -5,12 +5,12 @@ import hashlib
 from inspect import isgenerator
 from os import chmod, close, makedirs, remove
 from os.path import dirname, exists
-from passlib.hash import apr_md5_crypt
 from random import shuffle
 import stat
 from sys import stderr, stdout
 from tempfile import mkstemp
 
+from passlib.hash import apr_md5_crypt
 from requests import get
 
 from ..exceptions import DontCache, FaultUnavailable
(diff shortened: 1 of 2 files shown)


bundlewrap/utils/__init__.py
```diff
@@ -5,6 +5,7 @@
 from inspect import isgenerator
 from os import chmod, close, makedirs, remove
 from os.path import dirname, exists
+from passlib.hash import apr_md5_crypt
```
trehn, line 8:
```suggestion
```

trehn: Please move that down to where requests is imported from passlib.hash import apr_md5_crypt

PR #581, opened 2021-01-15
bw test: Merge "-d" and "-m"

vain:
Metadata is only calculated once now and then we check if metadata or
cdicts have changed.

Before:

    $ time bw test -d 2 -m 2 my.dns.server
    ✓ Metadata remained the same after being generated 2 times
    ✓ Configuration remained the same after being generated 2 times

    real    0m48.516s
    user    0m47.252s
    sys     0m1.251s

After:

    $ time bw test -d 2 -m 2 my.dns.server
    ✓ Configuration remained the same after being generated 2 times
    ✓ Metadata remained the same after being generated 2 times

    real    0m35.822s
    user    0m34.920s
    sys     0m0.914s

### DIFF
--- a/bundlewrap/cmdline/test.py
+++ b/bundlewrap/cmdline/test.py
@@ -133,93 +133,83 @@ def test_empty_groups(repo):
         exit(1)
 
 
-def test_determinism_config(repo, nodes, iterations, quiet):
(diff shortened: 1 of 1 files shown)


vain:
The performance boost isn’t as big as I hoped, but it’s something. :/

PR #586, opened 2021-01-22
User: Fix for FreeBSD

ThreeFx:
FreeBSD's user management is slightly different from the other OSes, this PR fixes this.

It's ad-hoc and (IMO) pretty ugly, but I believe the added complexity of refactoring the argument
dictionary to be OS-specific is too high (for now).

Any comments appreciated!

### DIFF
--- a/bundlewrap/items/users.py
+++ b/bundlewrap/items/users.py
@@ -108,7 +108,7 @@ class User(Item):
     @classmethod
     def block_concurrent(cls, node_os, node_os_version):
         # https://github.com/bundlewrap/bundlewrap/issues/367
-        if node_os == 'openbsd' or node_os == 'freebsd':
+        if node_os in ('openbsd', 'freebsd'):
             return [cls.ITEM_TYPE_NAME]
         else:
             return []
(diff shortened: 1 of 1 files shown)


bundlewrap/items/users.py
```diff
@@ -108,7 +108,7 @@ class User(Item):
     @classmethod
     def block_concurrent(cls, node_os, node_os_version):
         # https://github.com/bundlewrap/bundlewrap/issues/367
-        if node_os == 'openbsd':
+        if node_os == 'openbsd' or node_os == 'freebsd':
```
trehn, line 111:
```suggestion
        if node_os in ('freebsd', 'openbsd'):
```

bundlewrap/items/users.py
```diff
@@ -134,18 +134,33 @@ def cdict(self):
...
+                        # Using -H <n> we pass the password hash using file descriptor <n> instead.
+                        option = '-H'
+                        value = '0'
+                        command = "printf '%s' '{}' | ".format(self.attributes[attr]) + command
                     else:
                         value = str(self.attributes[attr])
                     command += "{} {} ".format(option, quote(value))
-            command += self.name
+            if self.node.os != 'freebsd':
+                command += self.name
```
trehn, line 163: Let's just do it that way for Linux as well, `usermod trehn -c test` works just fine.

bundlewrap/items/users.py
```diff
@@ -134,18 +134,33 @@ def cdict(self):
...
                 if (attr in status.keys_to_fix or status.must_be_created) and \
                         self.attributes[attr] is not None:
                     if attr == 'groups':
                         value = ",".join(self.attributes[attr])
+                    elif attr == 'password_hash' and self.node.os == 'freebsd':
+                        # On FreeBSD, pw useradd/usermod -p sets the password expiry time.
+                        # Using -H <n> we pass the password hash using file descriptor <n> instead.
+                        option = '-H'
+                        value = '0'
+                        command = "printf '%s' '{}' | ".format(self.attributes[attr]) + command
```
trehn, line 158: For extra credit, you could make use of the `data_stdin` argument to `self.run`, that way we could keep the hash out of logs/ps.

trehn (commented):
Thanks for the PR :)

bundlewrap/items/users.py
ThreeFx: Extra credit achieved :smile:

ThreeFx:
It seems that I accidentally included the pre-review state of this PR in the `FreeBSDPkg` merge, sorry about that @trehn!

I've rebased this on the new master to ensure that the correct code is merged.

PR #587, opened 2021-01-23
Display newly created files

### LINKED ISSUE #410: Show diff for new files?
Currently, the user just gets a `will be created` message for newly created files. Is this really a good idea? Why hide the content?

Hiding the content can kill the workflow and leave admins wondering: "Okay, a new file, but what's in it? I'd like to know. Now I'll have to check manually."

I'm aware that `status.must_be_created` is used in a lot of places ... This change request might not be as trivial as it seems.

ThreeFx:
Leverage the existing functionality to display a file diff on newly
created files. User, group and mode of the created file is displayed as
well.

This would fix #410.

Special-casing `file` in `Item/apply` is _really_ ugly, what do the maintainers think? I completely understand if you consider this unmergeable. In that case: how can we improve this? I'm not knowledgeable enough (just started using bundlewrap) to know more about the implementation.

Example output:

```
? test  ╭─ file:/etc/ssh/sshd_config
? test  │
? test  │  owner   → root
? test  │
? test  │  group   → root
? test  │
? test  │  mode   → 0644
? test  │
? test  │  content
? test  │  @@ -0,0 +1,29 @@
? test  │  +# This is the sshd server system-wide configuration file.  See
? test  │  +# sshd_config(5) for more information.
? test  │  +
? test  │  +# The configuration below was generated according to Mozilla's modern OpenSSH
? test  │  +# config: https://infosec.mozilla.org/guidelines/openssh.html
? test  │  +
? test  │  +# Listed in order of preference
? test  │  +HostKey /etc/ssh/ssh_host_ed25519_key
? test  │  +HostKey /etc/ssh/ssh_host_rsa_key
? test  │  +HostKey /etc/ssh/ssh_host_ecdsa_key
? test  │  +
? test  │  +# Only use secure Key Exchange/Encryption algorithms
? test  │  +KexAlgorithms curve25519-sha256@libssh.org,ecdh-sha2-nistp521,ecdh-sha2-nistp384,ecdh-sha2-nistp256,diffie-hellman-group-exchange-sha256
? test  │  +Ciphers chacha20-poly1305@openssh.com,aes256-gcm@openssh.com,aes128-gcm@openssh.com
? test  │  +MACs hmac-sha2-512-etm@openssh.com,hmac-sha2-256-etm@openssh.com,umac-128-etm@openssh.com,hmac-sha2-512,hmac-sha2-256,umac-128@openssh.com
? test  │  +
? test  │  +# Disable root login, disable password authentication
? test  │  +AuthenticationMethods publickey
? test  │  +PermitRootLogin no
? test  │  +
? test  │  +# Log key fingerprints of users
? test  │  +LogLevel VERBOSE
? test  │  +
? test  │  +# Log sftp level file access (read/write/etc.) that would not be easily logged otherwise.
? test  │  +Subsystem sftp  /usr/lib/ssh/sftp-server -f AUTHPRIV -l INFO
? test  │  +# End Mozilla recommendations
? test  │  +
? test  │  +# Allow client to pass locale environment variables
? test  │  +AcceptEnv LANG LC_*
? test  │
? test  ╰─ Fix file:/etc/ssh/sshd_config? [Y/n] n
```

### DIFF
--- a/bundlewrap/items/__init__.py
+++ b/bundlewrap/items/__init__.py
@@ -532,7 +532,12 @@ class Item:
                 )):
                     self.fix(status_before)
             else:
-                if status_before.must_be_created:
+                # In order to display new files which are created we have to
+                # ignore the default `must_be_created` behaviour for files.
+                #
+                # See https://github.com/bundlewrap/bundlewrap/pull/587 for
+                # more information.
+                if status_before.must_be_created and self.ITEM_TYPE_NAME != 'file':
                     question_text = _("Doesn't exist. Will be created.")
                 elif status_before.must_be_deleted:
                     question_text = _("Found on node. Will be removed.")
--- a/bundlewrap/items/files.py
+++ b/bundlewrap/items/files.py
@@ -341,19 +341,31 @@ class File(Item):
             }
 
     def display_dicts(self, cdict, sdict, keys):
-        if (
-            'content_hash' in keys and
-            self.attributes['content_type'] not in ('base64', 'binary') and
-            sdict['size'] < DIFF_MAX_FILE_SIZE and
-            len(self.content) < DIFF_MAX_FILE_SIZE and
-            PathInfo(self.node, self.name).is_text_file
-        ):
-            keys.remove('content_hash')
-            keys.append('content')
-            del cdict['content_hash']
-            del sdict['content_hash']
-            cdict['content'] = self.content
-            sdict['content'] = get_remote_file_contents(self.node, self.name)
+        if sdict is None:  # The target file does not exist
+            if (
+                self.attributes['content_type'] not in ('base64', 'binary') and
+                len(self.content) < DIFF_MAX_FILE_SIZE
+            ):
+                keys = ['owner', 'group', 'mode', 'content']
+                with suppress(KeyError):  # happens for content_type any
+                    del cdict['content_hash']
+                cdict['content'] = self.content
+                sdict = {'content': b'', 'mode': '', 'owner': '', 'group': ''}
+        else:
+            if (
+                'content_hash' in keys and
+                self.attributes['content_type'] not in ('base64', 'binary') and
+                sdict['size'] < DIFF_MAX_FILE_SIZE and
+                len(self.content) < DIFF_MAX_FILE_SIZE and
+                PathInfo(self.node, self.name).is_text_file
+            ):
+                keys.remove('content_hash')
+                keys.append('content')
+                del cdict['content_hash']
+                del sdict['content_hash']
+                cdict['content'] = self.content
+                sdict['content'] = get_remote_file_contents(
+                    self.node, self.name)
         if 'type' in keys:
             with suppress(ValueError):
                 keys.remove('content_hash')
--- a/bundlewrap/items/files.py
+++ b/bundlewrap/items/files.py
@@ -341,7 +341,7 @@ class File(Item):
             }
 
     def display_dicts(self, cdict, sdict, keys):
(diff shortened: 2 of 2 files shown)


bundlewrap/items/files.py
```diff
@@ -341,19 +341,31 @@ def sdict(self):
...
-            del sdict['content_hash']
-            cdict['content'] = self.content
-            sdict['content'] = get_remote_file_contents(self.node, self.name)
+        if sdict is None and not self.attributes.get('delete', False):
+            if (
+                self.attributes['content_type'] not in ('base64', 'binary') and
+                len(self.content) < DIFF_MAX_FILE_SIZE
+            ):
+                keys.extend(['owner', 'group', 'mode', 'content'])
+                with suppress(KeyError):
```
trehn, line 350:
```suggestion
                with suppress(KeyError):  # happens for content_type any
```

bundlewrap/items/files.py
```diff
@@ -341,19 +341,31 @@ def sdict(self):
...
-            del cdict['content_hash']
-            del sdict['content_hash']
-            cdict['content'] = self.content
-            sdict['content'] = get_remote_file_contents(self.node, self.name)
+        if sdict is None and not self.attributes.get('delete', False):
+            if (
+                self.attributes['content_type'] not in ('base64', 'binary') and
+                len(self.content) < DIFF_MAX_FILE_SIZE
+            ):
+                keys.extend(['owner', 'group', 'mode', 'content'])
```
trehn, line 349: Why are you _extending_ `keys` here? Is there any point in preserving what's in there?

```suggestion
                keys = ['owner', 'group', 'mode', 'content']
```

bundlewrap/items/__init__.py
```diff
@@ -532,7 +532,7 @@ def apply(
                 )):
                     self.fix(status_before)
             else:
-                if status_before.must_be_created:
+                if status_before.must_be_created and self.ITEM_TYPE_NAME != 'file':
```
trehn, line 535: I think special-casing this is the way to go and preferable to creating some complicated abstraction for this case that no other item is going to use 👍🏻

bundlewrap/items/files.py
ThreeFx: Mainly because I wasn't sure what was passed from above and wanted to be compatible.

I agree that it makes more sense to overwrite it.

ThreeFx:
I've also added a comment in `items/__init__.py` explaining why we special case `file`s and a link to this PR.

ThreeFx:
Should we maybe also display "this is a new file" somewhere? Currently, the user cannot distinguish between a diff with an empty file and a diff of a new file.

Maybe add `exists No -> Yes` or similar?

trehn:
I was wondering about this, too. Currently pondering if we shouldn't instead just do this for every item by just looking at the `cdict` returned by `display_dicts`...

ThreeFx:
I can see the merit of doing that, but I don't have enough experience with BundleWrap to think about the consequences.

Basically this should be possible by

1. setting `{cdict/sdict}['exists'}` based on the `status.must_be_{created,deleted}` variables
2. unifying the logic for asking questions in `apply()`

In that case we would even get rid of the special case `file`, hmm... Does `exists` make sense for every item?

Alternatively we could have each model handle this by itself, which would require a lot of code changes.

Depending on the solution it would make a lot of sense to do this in a separate PR.

trehn:
Forgive me for rejecting this, but I have created a generic version of this PR in #594 that works for all items. Many thanks for getting the ball rolling on this though!

PR #589, opened 2021-01-25
Group: Fix module on FreeBSD installs

ThreeFx:
As is the case with users, groups are managed by pw(8), and
groups cannot be managed concurrently.

IMO this also cleans up the `fix()` method a little bit.


trehn:
Thanks!

Issue #530, opened 2020-07-17
`bw repo create`: Do not clobber existing repos


vain:
For reasons, I had a dangling `$BW_REPO_PATH` in my environment that pointed to some existing repo. Then I ran `bw repo create` in `/tmp/foo`. It happily nuked the files in that other repo pointed to by the env var.

`bw repo create` should first check if any of the files it wants to create already exist.

PR #592, opened 2021-02-13
bw repo create: Do not clobber existing repos

vain:
Do as the comment says and abort if the directory is not empty.

CC #530 

Note that both the old and the new code has a race condition. If two processes try to do this at the same time, weird things will happen. I can only see one clean solution to this: Stop requiring that the directory already exists. Then we can do an `mkdir()` and the one process where this succeeds wins and creates a consistent new repo. Same idea as with directory-based locks in the file system.

That race condition is probably pretty rare, though, and I’m not sure if fixing it is worth breaking the API.


trehn:
thanks!

PR #595, opened 2021-02-16
Allow checking `action`s against multiple expected return codes

cmur2:
I propose allowing to pass in a list of integers to the `expected_return_code` attribute of an `action` in addition to passing in a single integer. The meaning here is that any of the listed integers signifies a successful execution of the action's `command`.

One use case I have is running `apt-get update` which may either exit with `0` (package lists got updated) or `100` (another process held the lock on the package lists) - in my setup I'd be fine with both as I just want to at least try to update the package lists once on a run. Currently I have to ignore the exit code alltogether on that aciton.

Maybe similiar use cases exist as well, where you want to allow certain exit codes without ignoring all of them.

I also added some tests for the `expected_return_code` attribute in general and my proposed addition of allowing lists in there. In case you like the direction of the PR but have concerns about the code quality or tests, please advise on desired changes :)

### DIFF
--- a/bundlewrap/items/actions.py
+++ b/bundlewrap/items/actions.py
@@ -186,9 +186,13 @@ class Action(Item):
                 may_fail=True,
             )
 
-        if self.attributes['expected_return_code'] is not None and \
-                not result.return_code == self.attributes['expected_return_code']:
-            raise ActionFailure(_("wrong return code: {}").format(result.return_code))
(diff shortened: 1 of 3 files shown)


trehn:
Could you update the docs as well?

cmur2:
@trehn thank you for the hint, I adjusted the PR :+1:

trehn:
Thanks!

PR #596, opened 2021-02-19
Add support for Alpine Linux

gjabell:
This patchset implements Alpine Linux support to bundlewrap by adding two new items:
- `pkg_apk`: Item for interacting with Alpine's `apk` package manager. Currently supports installing and uninstalling packages.
- `svc_openrc`: Item for managing services using Alpine's default init system, `OpenRC`. Supports (re)starting, reloading, stopping, and enabling/disabling services.

I haven't written any tests since there doesn't seem to be a system in place for testing OS-specific features, but I'm happy to add those if anyone has a suggestion on how to do that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,5 @@
 /dist/
 /docs/build/
+
+*.egg-info
+__pycache__/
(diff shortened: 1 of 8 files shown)


trehn:
Very nice, thank you!

PR #598, opened 2021-02-20
Fix tests to use new repo branch name main

cmur2:
I saw that https://github.com/bundlewrap/bundlewrap was adjusted to use `main` branch instead of `master` in the last 1-2 days. As this branch name is relied on in tests of `git_deploy` item, they were failing. This simple PR proposes to fix this.


trehn:
thanks!

Issue #568, opened 2020-11-27
prevent tags from causing redundant dependency errors


trehn:
```python
actions['wait_online_for_setup'] = {
    'triggered': True,
    'needed_by':   [
        'tag:setup.target',
    ],
}
actions['setup'] = {
    'preceded_by': [
        'action:wait_online_for_setup',
    ],
    'tags': [
        'setup.target',
    ],
}
```

results in

```redundant dependency of action:setup in bundle 'foo' on action:wait_online_for_setup```

trehn:
fixed as of 76c7325b59bccd6109b9873bc08f3c08afbc535f

PR #599, opened 2021-02-21
inherit tags for canned_actions

veecue:
Thanks to canned action, it is very easy to create dependency errors
with code like this:

```
svc_systemd = {
    'test.service': {
        'tags': {'a'},
        'needed_by': {'!tag:a'},
    },
}
```

Therefore canned actions should inherit the tags of their parent
actions to avoid confisions by using the `!tag:foobar` selector.

### DIFF
--- a/bundlewrap/deps.py
+++ b/bundlewrap/deps.py
@@ -154,7 +154,10 @@ def _inject_canned_actions(items):
     for item in items:
         for canned_action_name, canned_action_attrs in item.get_canned_actions().items():
             canned_action_id = f"{item.id}:{canned_action_name}"
-            canned_action_attrs.update({'triggered': True})
+            canned_action_attrs.update({
+                'triggered': True,
+                'tags': item.tags,
+            })
             action = Action(
                 item.bundle,
                 canned_action_id,


trehn:
@CroneKorkN do you agree?

trehn:
related to #568

CroneKorkN:
I actually have a usecase, where i need to tag each action independently: there are tags representing stages of app deployent, like `install`, `start`, `setup`, `stop` or `restart`. When we need to do something after an app has started, but before it was stopped, we do something like
```
action['setup_app'] = {
  'action': 'some setup stuff',
  'needs': ['tag:app-start.target'],
  'needed_by': ['tag:app-stop.target'],
}
```

Maybe this could be resolved via an explicit `svc_systemd:test:start` canned action, wich is triggered by default. Then tags on the base item could be inherited by canned actions und still each action could be tagged individually, if needed.

veecue:
I don't see a conflict between this PR and @CroneKorkN 's suggestion. Inheriting tags could still be the default behaviour that could be made overwriteable in a future commit/PR

trehn:
The problem is that this isn't backwards-compatible. Right now we actually have bundles that depend on a tag which is assigned to a service and world break if that dependency were to suddenly include the restart action as well.

Nevertheless, I still agree that the behavior implemented in this PR is probably the right way to go, so I'm tagging it for 5.0.

veecue:
Thanks for the reply. I guess I will use a patched `bundlewrap` version until then.

Issue #553, opened 2020-10-01
Stages


trehn:
Stages introduce a new type of configurable dummy item:

```python
# items.py

stages = {
    'my_first_stage': {
    },
    'my_second_stage': {
        'needs': ['stage:my_first_stage'],
    },
}

files = {
    '/foo-1': {
        'stage': 'my_first_stage',
    },
    '/foo-2': {
        'stage': 'my_first_stage',
    },
}

actions = {
    'compress': {
        'command': 'gzip /foo-*'
        'stage': 'my_second_stage',
    },
}
```

Stages only allow the `needs`, `needed_by`, and `tags` item attributes. They work like 'item groups' and will add those values to the respective ones set on each item in a stage. Internal dummy items (`bundle:*`, `tag:*`, ...) are never part of any stage.

BundleWrap creates three stages by default: first, default, last. All items default to the 'default' stage. All user-defined stages and the 'default' stage get an implicit dependency on 'first'. 'last' gets an implicit dependency on all user-defined stages and the default stage.

Stages other than `default` will log a message during `bw apply` when they start and end (as long as they have items).

trehn:
I'm also considering some internal refactoring to get rid of DummyItems. All these subclasses are really just ways to address groups of items. Having them as items in the dependency tree was just a hack that seemed easy before `needed_by` and such came around.

Since stages really are just another kind of item group, I don't really want them in `items.py`. This is also evident in that they wouldn't take all the attributes regular items understand. So maybe this is actually the time to finally introduce a `bundle.py` where we can define stages and perhaps other things in the future.

veecue:
This seems to be a much more mature concept that what I came up with. And always +1 from my side for allowing more metaprogramming :) (I LOVE metadata reactors already)

trehn:
As of #559, stages can now be realized through tags and the new bundle.py:

```python
# bundle.py next to items.py
tags = {
    'before-everything-else': {
        'needed_by': {'tag:', '!tag:'},  # all items with *other* tags and all untagged items
    },
    'normal-stage-1': {},
    'normal-stage-2': {
        'needs': {'tag:normal-stage-1'},
    },
    'normal-stage-3': {
        'needs': {'tag:normal-stage-2'},
    },
    'after-everything-else': {
        'needs': {'tag:', '!tag:'},  # all items with *other* tags and all untagged items
    },
}
```

veecue:
Thanks for the implementation.
[...]

PR #600, opened 2021-02-21
poc: Allow adding dependencies that should not be skipped

veecue:
Especially for stages (#553) it makes sense to be able to disable cascade skip for certain  dependencies (in this case, those specified in `bundle.py`). This poc shows how this could be achieved using magic strings.

Example `bundle.py` for working stages:
```python
tags = {
    # Everything that should be done before all normal steps
    'pre-all': {
        'needed_by': {'!tag:pre-all:noskip'},  # all items without this tag
    },
    'init': {
        'tags': {'pre-all'},
        'needed_by': {'tag:update-system:noskip'},
    },
    'update-system': {
        'tags': {'pre-all'},
    },

    # Everything that should be done after all normal steps
    'post-all': {
        'needs': {'!tag:post-all:noskip'},  # all items without this tag
    },
}
```
Open points for discussion:
 - [ ] Should this just be the default for dependencies specified on the tag level? It seems like the more intuitive behaviour since an action passing the `unless` condition is already a skip and might cause undesired cascade skips
 - [ ] What is a nice notation for this? The magic string `:noskip` won't do imo.
 - [ ] How can this be implemented nicely? The current implementation always skips if there is a `noskip` dependency, no matter if there is also a `skip` dependency which should take precedence.
 - [ ] How can such mostly tag-level dependencies be plotted (`bw plot node`) without clobbering the whole plot?

### DIFF
--- a/bundlewrap/deps.py
+++ b/bundlewrap/deps.py
@@ -131,9 +131,19 @@ def _has_trigger_path(items, item, target_item_id):
 def _prepare_deps(items):
     for item in items:
         item._deps = set()
+        item._noskip_deps = set()
+
         for dep in list(item.needs) + list(item.get_auto_deps(items)):
             try:
-                item._deps.update(resolve_selector(dep, items, originating_item_id=item.id))
+                noskip = False
+                if dep.endswith(":noskip"):
+                    dep = dep[:-7]
+                    noskip = True
+
+                deps = set(resolve_selector(dep, items, originating_item_id=item.id))
+                item._deps.update(deps)
+                if noskip:
+                    item._noskip_deps.update(map(lambda item: item.id, deps))
             except NoSuchItem:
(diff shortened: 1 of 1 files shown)


trehn:
I have tried to avoid this, but complexity has a way of catching up with you: Maybe it's time to introduce `after` (similar to the systemd meaning) in addition to `needs`. Meaning a dependency only affecting execution order, but not indicating an actual requirement for the item to be completed successfully. That would make `cascade_skip` mostly obsolete.

veecue:
Seems like a reasonable thing to do. I think most of the time when people use dependencies, they want the dependent item to be executed after the dependency but only if the dependency does not fail. For my use case that could be solved by just adding a flag somewhere to fail the whole apply operation immediately after the first error.

trehn:
should be good now with #609

PR #601, opened 2021-02-21
Speedup CI by using SSH multiplexing on localhost

cmur2:
Reduces CI runtime on GHA from ~10minutes ([old run](https://github.com/bundlewrap/bundlewrap/actions/runs/580929111)) to ~3minutes, providing faster feedback and avoids wasting resources.

**Note:** this PR also adjusts the GHA workflow to use `main` branch name!

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,9 +2,9 @@ name: Tests
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


trehn:
Nice catch, thanks!

PR #607, opened 2021-02-27
Remove redundant block_concurrent() from pkg_opkg, is already defined in Pkg

cmur2:
As `OpkgPkg` is derived from `Pkg` which already defines `block_concurrent()` in [a generic way](https://github.com/bundlewrap/bundlewrap/blob/main/bundlewrap/items/pkg.py#L18-L20), the re-definition with same functionality in `OpkgPkg` can be removed.


trehn:
thanks

PR #613, opened 2021-03-05
Don't rely on sudo to run postgres items

### LINKED ISSUE #612: Postgres operations require sudo
So far I haven't had any issues using `doas` instead of `sudo` by setting the `cmd_wrapper_outer` option. Unfortunately, the `postgres_role` and `postgres_db` items hardcode sudo to run commands as the postgres user, which breaks on a system without sudo installed. It would be nice if there was a way to specify which "privilege escalator" to use for commands such as these.

gjabell:
This patch adds a new group/node attribute called `run_as_command`, which specifies the command that should be used to run commands as another user. This defaults to `sudo -u {user} {command}` but can easily be changed to allow other privilege escalators, for example: `doas -u {user} {command}`. This functionality is wrapped in a new Node method, `node.run_as(user, command, ...)`.

The postgres items have been updated to use the new `run_as` method, as well as using psql's `-c` argument rather than piping SQL directly into psql, as this makes it easier to run under the correct user.

Also updated the Postgres docs to include the `delete` attribute, which seems to be missing.

Fixes #612

### DIFF
--- a/bundlewrap/group.py
+++ b/bundlewrap/group.py
@@ -19,7 +19,7 @@ from .utils.text import mark_for_translation as _, toml_clean, validate_name
 
 GROUP_ATTR_DEFAULTS = {
     'cmd_wrapper_inner': "export LANG=C; {}",
-    'cmd_wrapper_outer': "sudo sh -c {}",
+    'cmd_wrapper_outer': "sudo -u {1} sh -c {0}",
     'lock_dir': "/var/lib/bundlewrap",
     'dummy': False,
     'kubectl_context': None,
--- a/bundlewrap/items/postgres_dbs.py
+++ b/bundlewrap/items/postgres_dbs.py
@@ -7,7 +7,7 @@ from bundlewrap.utils.text import force_text, mark_for_translation as _
 
 def create_db(node, name, owner, when_creating):
     template = None
-    cmd = "sudo -u postgres createdb -wO {} ".format(owner)
+    cmd = "createdb -wO {} ".format(owner)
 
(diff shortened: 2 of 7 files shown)


cmur2:
I also have a usecase for such a `run_as_command` in a custom item I wrote so a functionality like this would be very welcome 🙂 

(My custom item needs to execute a script in the context of other users to create appropriate dotfiles configuration)

gjabell:
@cmur2 Nice :) please let me know if you have any suggestions for the patch; I'm thinking it might be good to extend this even further to intelligently recognize the type of privilege escalator so we don't have to specify both this and the `cmd_outer_wrapper`, but I'm not sure if that'd be too large a change.

trehn:
I don't like the redundancy with `cmd_wrapper_outer`. I'd also rather have an optional user argument to `node.run()` than a dedicated method.

The question is how to do this in a backwards-compatible way. We could have the username default to `root` and change the default of `cmd_wrapper_outer` to `sudo -u {1} sh -c {0}` (passing the username as the second argument aka `{1}`).

This will however break postgres items for people that have customized `cmd_wrapper_outer` (as the commands would no longer be run as the `postgres` user). But since these are probably those using `doas` anyway (and thus affected by #612), we could call it a fix.

gjabell:
@trehn That sounds like a better solution, thanks! I'll update the PR once I have some time this week.

gjabell:
Finally had some time to refactor this weekend; let me know if this new implementation looks better @trehn.

trehn:
Looks great, thank you!

PR #617, opened 2021-03-18
add localhost SSH verification steps

usrme:
Just to make it a little bit clearer on what to do exactly to get going quicker.


trehn:
thanks!

PR #629, opened 2021-05-22
Add external password providers

gjabell:
This PR adds the ability to load passwords or secrets from external providers, rather than relying on bundlewrap to manage secrets. I already use [pass](https://www.passwordstore.org/) for managing my passwords and server secrets, and having to manually reencrypt existing secrets (for example, user passwords which I can't have managed purely in bundlewrap) is tedious (plus I don't like how long the strings are in my templates :grimacing:).

The new `password_from` method accepts an identifier (ie the name of the password to load), as well as an optional external password provider. This provider should be a command that can be interpolated to include the identifier.

This patch also adds a new file called `repo.cfg` for storing repository-wide configuration values; in this case the list of external password providers as well as the default provider to use if none is specified. I'm not sure if this is the best place to add these config values, but it didn't seem right to add them to .secrets.cfg, so if there is a preferred place to add this configuration without creating a new file, I can refactor that.

### DIFF
--- a/bundlewrap/repo.py
+++ b/bundlewrap/repo.py
@@ -1,3 +1,4 @@
+from configparser import ConfigParser
 from contextlib import suppress
 from importlib.util import module_from_spec, spec_from_file_location
 from inspect import isabstract
@@ -40,6 +41,8 @@ DIRNAME_LIBS = "libs"
 FILENAME_GROUPS = "groups.py"
 FILENAME_NODES = "nodes.py"
 FILENAME_REQUIREMENTS = "requirements.txt"
+FILENAME_REPO_CONFIG = "repo.cfg"
+
 
 HOOK_EVENTS = (
     'action_run_end',
@@ -92,6 +95,11 @@ nodes = {
     """),
     FILENAME_REQUIREMENTS: "bundlewrap>={}\n".format(VERSION_STRING),
     FILENAME_SECRETS: generate_initial_secrets_cfg,
+    FILENAME_REPO_CONFIG: """
+[DEFAULT]
+
+[password_providers]
+"""
 }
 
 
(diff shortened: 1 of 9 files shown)


cmur2:
I'm also interested in a `pass` integration as I currently use a custom library function like this:

```python
def get_pass_password(pass_path: str) -> str:
    if os.environ.get('BW_VAULT_DUMMY_MODE', '0') != '0':
        return 'pass password'

    try:
        return subprocess.run(['pass', 'show', pass_path], stdout=subprocess.PIPE,
                              check=True).stdout.decode('utf-8').split('\n')[0]
    except BaseException as ex:
        raise bundlewrap.exceptions.FaultUnavailable from ex
```

This works well for my use cases and allows me to choose where I want to get a secret from based on which code I use: `repo.libs.my_custom_library.get_pass_password('path/to/secret')` or the `repo.vault.*` family of functions.

fkusei:
Have a look at https://git.kunsmann.eu/kunsi/bundlewrap-pass :) (not yet published to pypi, though)

gjabell:
@cmur2 @fkusei Very nice! This PR might be a bit redundant I guess :slightly_smiling_face: Although maybe it's still helpful to have something more generic in the base project?

trehn:
I'm not keen on introducing a `repo.cfg`. I also believe the benefit of having a pluggable provider system is rather limited if that only includes different shell commands (and not Python code).

So while I feel bad rejecting a nice PR with docs and tests and everything, I'd much prefer Franzis solution. For a more generic approach, I have also created https://github.com/bundlewrap/bundlewrap/pull/633.

gjabell:
No worries, thanks for the feedback :slightly_smiling_face: I'll use one of the suggested libs or the new vault cmd.

Kunsi:
If anyone is still interested: bundlewrap-pass was just uploaded to PyPI: https://pypi.org/project/bundlewrap-pass/

PR #634, opened 2021-05-30
Users: Username must come last

vain:
On GNU (and probably FreeBSD), the order doesn't matter.

On OpenBSD, creating or modifying users was broken.

-----

This was probably broken since 3f9bdf8fd749d4cba6ffc69fa3c562b6d80d68ce, I’m not sure. I don’t have a FreeBSD box ready, so it would be great, if you, @ThreeFx, could give this PR a spin and see if creating users still works on FreeBSD. Thanks!

### DIFF
--- a/bundlewrap/items/groups.py
+++ b/bundlewrap/items/groups.py
@@ -60,16 +60,16 @@ class Group(Item):
         else:
             command += "groupadd " if status.must_be_created else "groupmod "
 
+            if self.attributes['gid'] is not None:
+                command += "-g {} ".format(self.attributes['gid'])
+
             if self.node.os == 'freebsd':
                 # FreeBSD expects <name> to be the first argument to
                 # `pw groupadd/mod`, however we can also pass it using -n
                 # instead. Then it is positionally independent.
                 command += "-n "
 
-            command += f"{self.name} "
-
-            if self.attributes['gid'] is not None:
-                command += "-g {}".format(self.attributes['gid'])
+            command += f"{self.name}"
         self.run(command, may_fail=True)
 
     def sdict(self):
--- a/bundlewrap/items/groups.py
+++ b/bundlewrap/items/groups.py
@@ -59,6 +59,13 @@ class Group(Item):
             command += f"groupdel {self.name}"
         else:
             command += "groupadd " if status.must_be_created else "groupmod "
+
+            if self.node.os == 'freebsd':
+                # FreeBSD expects <name> to be the first argument to
+                # `pw groupadd/mod`, however we can also pass it using -n
+                # instead. Then it is positionally independent.
+                command += "-n "
+
             command += f"{self.name} "
 
             if self.attributes['gid'] is not None:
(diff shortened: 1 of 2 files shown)


ThreeFx:
> This was probably broken since 3f9bdf8

Whoops, sorry about that!

> @ThreeFx, could give this PR a spin and see if creating users still works on FreeBSD.

Good catch, it doesn't. But it works as `pw <operation> -n name/uid`. In that case it's positionally independent, so you can add it anywhere you want. Will you update the PR or should I send a patch?

vain:
It really is broken on FreeBSD now? 🙈

Yeah, would be nice if you could send a patch. :)  Hm, but maybe I should create a FreeBSD VM anyway, so I can see what’s going on there …

ThreeFx:
> It really is broken on FreeBSD now?

I must confess, I did not run the thing, but I tested the generated command:

    bfiedler@freebsd-host:~
    % doas pw groupadd -g 2345 testgroup
    pw: group name required

Switching it around:

    bfiedler@freebsd-host:~
    % doas pw groupadd testgroup -g 2346
    bfiedler@freebsd-host:~
    %

With `-n`:

    bfiedler@freebsd-host:~
    % doas pw groupadd -g 2345 -n testgroup
    bfiedler@freebsd-host:~
    %

So it should be an easy fix :)

ThreeFx:
```patch
diff --git a/bundlewrap/items/users.py b/bundlewrap/items/users.py
index 21050517..a8a0f4d7 100644
--- a/bundlewrap/items/users.py
+++ b/bundlewrap/items/users.py
@@ -152,8 +152,9 @@ class User(Item):
                     if attr == 'groups':
                         value = ",".join(self.attributes[attr])
                     elif attr == 'password_hash' and self.node.os == 'freebsd':
-                        # On FreeBSD, pw useradd/usermod -p sets the password expiry time.
-                        # Using -H <n> we pass the password hash using file descriptor <n> instead.
+                        # On FreeBSD, pw useradd/usermod -p sets the password
+                        # expiry time.  Using -H <n> we pass the password hash
+                        # using file descriptor <n> instead.
                         option = '-H'
                         value = '0'  # FD 0 = stdin
                         stdin = self.attributes[attr].encode()
@@ -161,6 +162,12 @@ class User(Item):
                         value = str(self.attributes[attr])
                     command += "{} {} ".format(option, quote(value))
 
+            if self.node.os == 'freebsd':
+                # FreeBSD expects <name> to be the first argument to
+                # `pw useradd/mod` , however we can also pass it using -n
+                # instead. Then it is positionally independent.
+                command += "-n "
+
             command += f"{self.name}"
             self.run(command, data_stdin=stdin, may_fail=True)
```

This is only for the user side though, have you checked whether this problem occurs in `bundlewrap/items/groups.py` as well?

vain:
@ThreeFx You were right, groups were broken, too. Turns out, I don’t manage any groups on my OpenBSD boxes, so I didn’t notice. 🤦 We should automate testing of this kind of stuff …

Anyway, thanks for the hints and the patch. I updated the PR. This is tested on OpenBSD 6.9, FreeBSD 13, and Ubuntu 20.04.

trehn:
Thanks!

PR #635, opened 2021-06-01
operations/upload: Preserve owner/group/mode of target if exists

### LINKED ISSUE #628: Should node.upload retain the permissions of the target file if it exists?
Currently I'm writing code for a new item which is something of a smart file manager (think of it as managing a handfull of files and having a more sophisticated replacement logic than "if different then update").

I'm using an auto dependency on all paths I'm managing: Each path must have a corresponding `file` definition. Idea here is that I'm not duplicating the file definition logic.

Now the problem arises when I want to use `node.upload` to upload the content into the (now correctly owned and permissioned) files. `mv` overwrites ownership (well not technically, it simply deletes the destination if it ever existed, but the effect is identical).

`cp` does not overwrite permissions by default. My question is whether it would make sense to change the semantic of `node.upload` to use `cp` followed by `rm` instead of `mv`.

Pros:
  * existing permissions are retained if the file exists

Cons
  * temporarily duplicates the used space

ThreeFx:
@vain helpfully pointed out that the current implementation conflicts
with the documentation, thanks!

As described in the bundlewrap documentation:

 > owner
 >
 > Username of the file's owner. Defaults to 'root'. Set to
 > None if you don't want BundleWrap to change whatever
 > is set on the node.

Setting owner to `None` and uploading a file causes the target to be
owned by the user bundlewrap uses to establish the ssh session, and
overwrites the previous owner of the target. The same principle applies
to group and mode.

`cp` by default does preserves the target permissions if it exists,
(which corresponds to the command line option --no-preserve=all, since
`cp`'s preservation applies to the source file).

Note on `cp -f`: GNU `cp` preserves ownership of the target even using
`cp -f`, but e.g. FreeBSD does not, hence why we use `cp` instead of
`cp -f`.

#### Further Discussion/TODO:

- [ ] Are the `cp` options okay like that? Should we be more explicit? I can offer to test this on Debian and FreeBSD.
- [ ] @vain do you have some OpenBSD machines to test this on?
- [ ] We still duplicate disk space, but IMO this is fine.

Closes #628

### DIFF
--- a/bundlewrap/operations.py
+++ b/bundlewrap/operations.py
@@ -330,9 +330,11 @@ def upload(
         if result.return_code != 0:
             return False
 
-    result = run(
+    cp_result = run(
         hostname,
-        "mv -f {} {}".format(
+        # cp preserves the target permissions by default. On Debian, cp -f
+        # also preserves ownership but this is not the case on e.g. FreeBSD
+        "cp {} {}".format(
             quote(temp_filename),
             quote(remote_path),
         ),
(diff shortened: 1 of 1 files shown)


ThreeFx:
Ah it gets better: This breaks the case when the target file does not exist, since cp will create a new file with non-inherited permissions...

ThreeFx:
The best I can do is `if [ -f remote_file ]; then cp tmpfile remote_file; else mv tmpfile remote_file; fi`.

ThreeFx:
Maybe it would be better to first copy the file from the temporary path to the final path and then setting it's mode/ownership. This however breaks the all-or-nothing guarantee: a failing execution may leave a file with wrong modes behind.

PR #640, opened 2021-07-02
bundlewrap.utils.Fault: add __iter__()

fkusei:


fkusei:
Well, that does not work as i expected it to work ...

PR #645, opened 2021-07-15
file: clarify the behaviour of `None` for owner, mode and group

ThreeFx:
- [ ] Not sure about `0644`, is this always the case?


trehn:
Thanks!

PR #651, opened 2021-07-21
metagen: Do not run everything again if no reactor ran

vain:
This saves quite some time on our big repo (`bw hash icinga2` from 5.5 minutes / 1.1 TC to 4 minutes / 0.8 TC, TC being “tera cycles” according to `perf stat`).

All the hashes of `bw hash` and `bw hash -m` commands that I ran remained the same, but I’m still running some more at the moment.

### DIFF
--- a/bundlewrap/metagen.py
+++ b/bundlewrap/metagen.py
@@ -285,6 +285,7 @@ class MetadataGenerator:
 
         self.__reset()
         self.__nodes_that_never_ran.add(initial_node_name)
+        self._some_reactor_ran = False
 
         while not QUIT_EVENT.is_set():
             jobmsg = _("{b} ({n} nodes, {r} reactors, {e} runs)").format(
@@ -317,7 +318,12 @@ class MetadataGenerator:
                     # dependencies on other nodes. If anything changes, we need to start over so
                     # local-only reactors on a node can react to changes caused by reactors looking
                     # at other nodes.
-                    self.__run_nodes()
+                    #
+                    # However, if no reactor ran during the steps above,
+                    # then there is no chance that something could have
+                    # changed.
+                    if self._some_reactor_ran:
+                        self.__run_nodes()
 
                     # If we get here, we're done! All that's left to do is blacklist completed
                     # reactors so they don't get run again if additional metadata is requested.
@@ -421,6 +427,7 @@ class MetadataGenerator:
             self._node_metadata_proxies[node.name]._satisfied = True
 
     def __run_reactor(self, node, reactor_name, reactor):
+        self._some_reactor_ran = True
         if (node.name, reactor_name) in self.__do_not_run_again:
             return False, set()
         self._partial_metadata_accessed_for = set()
--- a/bundlewrap/metagen.py
+++ b/bundlewrap/metagen.py
@@ -213,6 +213,10 @@ class MetadataGenerator:
     # should we collect information for `bw plot reactors`?
     _record_reactor_call_graph = False
 
+    # This only exists to make DeepSource happy
+    def __init__(self):
+        self._some_reactor_ran = False
+
     def __reset(self):
         # reactors that raised DoNotRunAgain
         self.__do_not_run_again = set()
@@ -240,6 +244,8 @@ class MetadataGenerator:
         self._additional_path_requested = False
         # bw plot reactors
         self._reactor_call_graph = set()
+        # To find out if we can skip a final "__run_nodes()"
+        self._some_reactor_ran = False
 
     def _metadata_proxy_for_node(self, node_name):
         if node_name not in self._node_metadata_proxies:
@@ -285,7 +291,6 @@ class MetadataGenerator:
 
         self.__reset()
         self.__nodes_that_never_ran.add(initial_node_name)
-        self._some_reactor_ran = False
 
         while not QUIT_EVENT.is_set():
             jobmsg = _("{b} ({n} nodes, {r} reactors, {e} runs)").format(
--- a/bundlewrap/repo.py
+++ b/bundlewrap/repo.py
@@ -197,6 +197,7 @@ class Repository(MetadataGenerator):
         # required by MetadataGenerator
         self._node_metadata_proxies = {}
         self._node_metadata_lock = RLock()
+        self._some_reactor_ran = False
 
         if repo_path is not None:
             self.populate_from_path(self.path)
--- a/bundlewrap/repo.py
+++ b/bundlewrap/repo.py
@@ -181,6 +181,8 @@ class LibsProxy:
 
 class Repository(MetadataGenerator):
     def __init__(self, repo_path=None):
+        super().__init__()
+
         if repo_path is None:
             self.path = "/dev/null"
         else:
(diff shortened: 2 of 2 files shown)


vain:
I’ll look at the DeepSource complaints once we have decided that we actually want to merge this …

vain:
The value of `bw hash` for our entire repo remains the same.

bundlewrap/metagen.py
```diff
@@ -213,6 +213,10 @@ class MetadataGenerator:
     # should we collect information for `bw plot reactors`?
     _record_reactor_call_graph = False
 
+    # This only exists to make DeepSource happy
```
trehn, line 216: Please don't do this. Just let it complain.

bundlewrap/repo.py
```diff
@@ -181,6 +181,8 @@ def __getattr__(self, attrname):
 
 class Repository(MetadataGenerator):
     def __init__(self, repo_path=None):
+        super().__init__()
+
```
trehn, line 185: But is certainly more correct this way.

Should also move
```python
        # required by MetadataGenerator
        self._node_metadata_proxies = {}
        self._node_metadata_lock = RLock()
```
into `MetadataGenerator.__init__` then.

trehn:
I don't get why this saves time. As soon as any reactor is involved, it shouldn't trigger. 🤔

vain:
> I don't get why this saves time. As soon as any reactor is involved, it shouldn't trigger.

Suppose you have this bundle:

```python
files['/etc/foo'] = {
    'content': node.metadata.get('something', 'whatever'),
}

files['/etc/bar'] = {
    'content': node.metadata.get('something_else', 'meh'),
}
```

Creating all items results in this log:
```
    [2021-07-22T16:40:25.621101] [DEBUG] invocation: /home/klausdieter/git/bundlewrap/virtualenv/bin/bw items -w /tmp/tmp/stuff node-1
    [2021-07-22T16:40:25.688712] [DEBUG] node node-1 gets its dummy attribute from: default
    [2021-07-22T16:40:25.689028] [DEBUG] node-1 no longer satisfied because of ('something',)
    [2021-07-22T16:40:25.689093] [DEBUG] initial metadata run for node-1
    [2021-07-22T16:40:25.689385] [DEBUG] node-1:metadata_reactor:foo.something changed=True deps=set()
    [2021-07-22T16:40:25.689482] [DEBUG] running reactors for node-1
    [2021-07-22T16:40:25.689512] [DEBUG] node-1:metadata_reactor:foo.something changed=False deps=set()
(1) [2021-07-22T16:40:25.689564] [DEBUG] metadata generation for selected nodes finished
(2) [2021-07-22T16:40:25.689593] [DEBUG] node-1 no longer satisfied because of ('something_else',)
    [2021-07-22T16:40:25.689644] [DEBUG] initial metadata run for node-1
(3) [2021-07-22T16:40:25.689761] [DEBUG] running reactors for node-1
    [2021-07-22T16:40:25.689809] [DEBUG] metadata generation for selected nodes finished
    [2021-07-22T16:40:25.689913] [DEBUG] node node-1 gets its os attribute from: default
    [2021-07-22T16:40:25.690135] ✓ wrote /tmp/tmp/stuff/etc/bar
    [2021-07-22T16:40:25.690205] ✓ wrote /tmp/tmp/stuff/etc/foo
```
The lines with “no longer satisfied” come from this patch:
```diff
diff --git a/bundlewrap/metagen.py b/bundlewrap/metagen.py
index 59d3c9d8..518396df 100644
--- a/bundlewrap/metagen.py
+++ b/bundlewrap/metagen.py
@@ -170,6 +170,7 @@ class NodeMetadataProxy:
             # It needs to be an RLock because this method will be recursively
             # called from _build_node_metadata (when reactors call node.metadata.get()).
             if self._requested_paths.add(path):
+                io.debug(f'{self._node.name} no longer satisfied because of {path}')
                 self._satisfied = False
                 self.__relevant_reactors_cache = None
                 if self._metagen._in_a_reactor:
```
Now, what we see:

1.  We’re done with metadata generation – for the first item. The item can be written to disk now.
2.  The second item requests a totally new metadata path, thus marking the node as “no longer satisfied”.
3.  We have entered a whole new iteration of `_build_node_metadata()` and we have now reached `__run_nodes()` again. It won’t run any reactors, but it will still check if it has to. That’s the point where we can optimize the process: If neither `__run_new_nodes()` nor `__run_triggered_nodes()` ran *any* reactor, then there’s no point in running `__run_nodes()` now – nothing has changed. `__run_nodes()` does the same check internally anyway, my PR simply introduced a single boolean flag to make this check faster.

I saw this happening a lot in our big repo, at least with the test cases I happened to run. It doesn’t trigger with `bw test`, because that requests all metadata at once (`.get(tuple())` or something like that, I saw a line like `node-1 no longer satisfied because of ()`), but it does trigger with `bw hash` and `bw hash -m` and I *assume* it also triggers with `bw apply`.

bundlewrap/repo.py
vain: Will fix tomorrow.

vain:
I guess this mostly helps with metadata that is not set on a node, like `metadata.get('foo', 'this_default_is_used')` or `if metadata.get('foo', None): do_something()`. We probably have *a ton* of that?

trehn:
I think this breaks when a reactors accesses metadata from a new node that doesn't have any reactors.

Initially, the reactor will encounter a `KeyError` because the metastack for that other node is empty. Then `__run_new_nodes` adds static metadata to that metastack, but that doesn't add the original node to `__triggered_nodes`, so `__run_triggered_nodes` won't do anything. Thus, no reactor ran and `__run_nodes` is skipped, never letting the original reactor see the new static metadata.

vain:
Hmm. I couldn’t verify that, yet. Did I miss anything? 🤔

Let there be two nodes, `A` and `B`. `A` has a reactor that requests data from `B`, that’s the reactor that might raise `KeyError`.

### Let’s say the new node is discovered by a metadata reactor

The flag `_some_reactor_ran` is only reset in `__reset()`. So the idea is, whenever we enter `_build_node_metadata()` afresh and no reactor whatsoever runs in `__run_new_nodes()` or `__run_triggered_nodes()`, then and only then will the flag stay at `False`, indicating that we can skip `__run_nodes()`.

In your example, the reactor on `A` ran and raised a `KeyError`. The fact that it ran sets `_some_reactor_ran` to `True` and it will stay that way for all the remaining iterations of the main `while` loop in `_build_node_metadata()`. Hence the optimization in this PR doesn’t matter anymore.

### Let’s say the new node is discovered by a template (or anything else outside the original reactor on `A`)

Let’s assume there’s a template that requests metadata of `B` and let’s assume that `B` was never looked at before. That *would* trigger the reactor on `A`, yes. But: Does `A` matter?

To get into this situation, the reactor on `A` never ran before – because that would have immediately pulled in `B`. So, `A` is totally irrelevant to the template we’re currently rendering.

At some later point, the reactor of `A` might indeed run. But then, the (static) metadata of `B` is already finished and `A` can just read it. This shouldn’t even raise a `KeyError` now.

### All this is way too complicated

Unless you’re going “ah, sure, right, let’s do this!” now, I don’t think we should investigate this optimization any further. In a few weeks, we both will have forgotten how this works, even if it’s just a simple flag.

Instead, we should investigate whole new approaches for generating metadata.

vain:
I guess this is obsolete, fup2 https://github.com/bundlewrap/bundlewrap/pull/655

PR #653, opened 2021-07-21
Use isoformat() in debug log

vain:
strftime() is expensive and isoformat() shows almost the same
information, the difference being that annoying "T" from ISO8601.

For large log output (couple of hundred MB), this can shave off a few
billion CPU cycles. It's not a lot, but it's something.

### DIFF
--- a/bundlewrap/utils/ui.py
+++ b/bundlewrap/utils/ui.py
@@ -43,7 +43,7 @@ def add_debug_timestamp(f):
     @wraps(f)
     def wrapped(self, msg, **kwargs):
         if self.debug_mode:
-            msg = datetime.now().strftime("[%Y-%m-%d %H:%M:%S.%f] ") + msg
+            msg = "[" + datetime.now().isoformat() + "] " + msg
         return f(self, msg, **kwargs)
     return wrapped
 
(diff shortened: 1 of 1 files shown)


bundlewrap/utils/ui.py
```diff
@@ -43,7 +43,7 @@ def add_debug_timestamp(f):
     @wraps(f)
     def wrapped(self, msg, **kwargs):
         if self.debug_mode:
-            msg = datetime.now().strftime("[%Y-%m-%d %H:%M:%S.%f] ") + msg
+            msg = "[" + datetime.now().isoformat() + "] " + msg
```
trehn, line 46: While we're micro-optimizing:

```suggestion
            msg = f"[{datetime.now().isoformat()}] {msg}"
```

This is actually faster.

bundlewrap/utils/ui.py
```diff
@@ -54,7 +54,7 @@ def wrapped(self, msg, **kwargs):
         if self.debug_log_file and self._active:
             with self.lock:
                 self.debug_log_file.write(
-                    datetime.now().strftime("[%Y-%m-%d %H:%M:%S.%f] ") +
+                    "[" + datetime.now().isoformat() + "] " +
```
trehn, line 57:
```suggestion
                    f"[{datetime.now().isoformat()}] {ansi_clean(msg).rstrip("\n")}\n"
```

bundlewrap/utils/ui.py
```diff
@@ -54,7 +54,7 @@ def wrapped(self, msg, **kwargs):
...
                     ansi_clean(msg).rstrip("\n") + "\n"
```
trehn, line 58:
```suggestion
```

bundlewrap/utils/ui.py
vain: Krchkrchkrch. `SyntaxError: f-string expression part cannot include a backslash`

I’ll push something similar.

PR #655, opened 2021-07-22
precise metadata reactor triggers

trehn:


trehn:
- [ ] unrandomize?
- [ ] better progress indication

Issue #370, opened 2017-09-26
`bw run --stdout-table` breaks table if output contains tabs


trehn:

trehn:
This also concerns other characters that aren't single-width.

vain:
FTR, test case:
```
$ bw run --stdout-table xxxxxxxxxxxxxxxxxxx 'printf "foo\tbar\n🐧\n"'
✓ xxxxxxxxxxxxxxxxxxx  bwlog  on_run_start  successful
› xxxxxxxxxxxxxxxxxxx  foo	bar
› xxxxxxxxxxxxxxxxxxx  🐧
✓ xxxxxxxxxxxxxxxxxxx  bwlog  on_run_end  successful
i ╭─────────────────────┬─────────────┬────────┬─────────╮
i │ node                │ return code │ time   │ stdout  │
i ├─────────────────────┼─────────────┼────────┼─────────┤
i │ xxxxxxxxxxxxxxxxxxx │           0 │ 0.068s │ foo	bar │
i │                     │             │        │ 🐧       │
i ╰─────────────────────┴─────────────┴────────┴─────────╯
```
This is going to be tough to fix properly, given all the Unicode shenanigans.

Do we think this is worth fixing? I dunno, never used this feature, tbh. 🫤

PR #662, opened 2021-08-03
Fix "job" lines being cut off too early

vain:
On narrow terminals, the status line showing the current job was being
cut off, because we counted escape sequences as visible characters.


vain:
Didn’t we have an open issue for this? I can’t find it.

trehn:
Maybe you're thinking of #370?

PR #666, opened 2021-08-24
items/file: add test_with

fkusei:
Same as `verify_with`, but called when running `bw test`.

### DIFF
--- a/bundlewrap/items/files.py
+++ b/bundlewrap/items/files.py
@@ -436,7 +436,7 @@ class File(Item):
                     exitcode, stdout = self._run_validator(cmd)
                     if exitcode == 0:
                         io.debug("{i} passed local validation".format(i=self.id))
-                    elif exitcode in (126,127,255):
+                    elif exitcode in (126, 127, 255):
                         io.debug("{i} failed local validation with code {c}, ignoring".format(i=self.id, c=exitcode))
                     else:
                         raise BundleError(_(
--- a/bundlewrap/items/files.py
+++ b/bundlewrap/items/files.py
@@ -4,7 +4,7 @@ from contextlib import contextmanager, suppress
 from datetime import datetime
 from os.path import basename, dirname, exists, join, normpath
 from shlex import quote
-from subprocess import call
+from subprocess import check_output, CalledProcessError, STDOUT
 from sys import exc_info
 from traceback import format_exception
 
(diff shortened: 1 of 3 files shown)


fkusei:
Maybe this should suppress output? I don't know

trehn:
Including the output of the command in the exception would be great.

I'm not sure how I feel about encouraging people using this for commands that won't be available to users locally. Maybe special-case return code 127 as a soft failure? I don't want to end up in a sitation where I can't run `bw test` locally anymore.

bundlewrap/items/files.py
```diff
@@ -429,8 +430,31 @@ def test(self):
...
         if not self.attributes['delete'] and not self.attributes['content_type'] == 'any':
-            with self._write_local_file():
-                pass
+            with self._write_local_file() as local_path:
+                if self.attributes['test_with']:
+                    cmd = self.attributes['test_with'].format(quote(local_path))
+                    exitcode, stdout = self._run_validator(cmd)
+                    if exitcode == 0:
+                        io.debug("{i} passed local validation".format(i=self.id))
+                    elif exitcode in (126,127,255):
```
fkusei, line 439: Same as in `operations.py` lines 200-202

trehn:
```suggestion
                    elif exitcode in (126, 127, 255):
```

docs/content/items/file.md
```diff
@@ -94,3 +94,10 @@ This can be used to run external validation commands on a file before it is appl
 <pre><code class="nohighlight">visudo -cf {}</code></pre>
 
 Keep in mind that all team members will have to have the verification command installed on their machines.
+<hr>
+
+## test_with
+
+Same as `verify_with`, but called when running `bw test`. You may want to use this if you don't want all your team members to have to have a large suite of tools installed, but still want to verify file integrity with another tool.
+
+The file is verified locally on the machine running BundleWrap. Verification is considered successful when the exit code of the verification command is 0. Use `{}` as a placeholder for the shell-quoted path to the temporary file.
```
trehn, line 103:
```suggestion
The file is verified locally on the machine running BundleWrap. Verification is considered successful if the exit code of the verification command is 0. Use `{}` as a placeholder for the shell-quoted path to the temporary file.
```

bundlewrap/items/files.py
```diff
@@ -429,8 +430,31 @@ def test(self):
...
-            with self._write_local_file():
-                pass
+            with self._write_local_file() as local_path:
+                if self.attributes['test_with']:
+                    cmd = self.attributes['test_with'].format(quote(local_path))
+                    exitcode, stdout = self._run_validator(cmd)
+                    if exitcode == 0:
+                        io.debug("{i} passed local validation".format(i=self.id))
+                    elif exitcode in (126, 127, 255):
+                        io.debug("{i} failed local validation with code {c}, ignoring".format(i=self.id, c=exitcode))
```
fkusei, line 440: Thinking about it, it would be nice if we could show that message to the user. That's similar to `FaultUnavailable` while rendering a template, *we could not test the file because of an error*.

PR #670, opened 2021-09-17
items/files: add content_type 'download'

Kunsi:
```
✘ rx300  gitea  file:/usr/local/bin/gitea  
│ 
│ content_hash
│   548018e2c9c1dfd9ac0919aec743759027ad9c72
│ → da39a3ee5e6b4b0d3255bfef95601890afd80709
│ source
│   <missing>
│ → https://dl.gitea.io/gitea/1.15.2/gitea-1.15.2-linux-amd64
╵
```

### DIFF
--- a/bundlewrap/items/files.py
+++ b/bundlewrap/items/files.py
@@ -11,6 +11,7 @@ from shutil import rmtree
 from subprocess import check_output, CalledProcessError, STDOUT
 from sys import exc_info
 from tempfile import gettempdir
+from time import sleep
 from traceback import format_exception
 
 from jinja2 import Environment, FileSystemLoader
@@ -20,9 +21,9 @@ from mako.template import Template
 from bundlewrap.exceptions import BundleError, FaultUnavailable, TemplateError
 from bundlewrap.items import BUILTIN_ITEM_ATTRIBUTES, Item
 from bundlewrap.items.directories import validator_mode
-from bundlewrap.utils import cached_property, hash_local_file, sha1, tempfile
+from bundlewrap.utils import cached_property, download, hash_local_file, sha1, tempfile
 from bundlewrap.utils.remote import PathInfo
-from bundlewrap.utils.text import force_text, mark_for_translation as _
+from bundlewrap.utils.text import bold, force_text, mark_for_translation as _
 from bundlewrap.utils.text import is_subdirectory
 from bundlewrap.utils.ui import io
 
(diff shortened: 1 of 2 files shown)


Kunsi:
I fixed 1 and implemented 3, but hashing local files with lots of megabytes is very slow :(

Kunsi:
Issues are fixed. 3) seems to be a local problem, it happens *sometimes* during running `bw verify`, but not always.

Not sure about what to do about 2) - There's no way to know where the file currenty on the node is from.

bundlewrap/items/files.py
```diff
@@ -133,9 +138,74 @@ def content_processor_text(item):
...
+            sleep(1)
+
+    verify_ssl = item.attributes.get('verify_ssl', True) # TODO
+
+    try:
+        if not isfile(file_path):
+            io.debug(_('{}: downloading from {}'.format(item.name, item.attributes['source'])))
+
+            item.run_local([
+                'curl',
```
trehn, line 180: Why use curl? We already depend on requests.

trehn: we even have a `bundlewrap.utils.download()`

bundlewrap/items/files.py
```diff
@@ -133,9 +138,74 @@ def content_processor_text(item):
...
+    else:
+        remove_dir = join(gettempdir(), 'bw-file-download-cache-{}'.format(getpid()))
+        file_path = join(remove_dir, file_name_hashed)
+        lock_dir = join(remove_dir, '{}.bw_lock'.format(file_name_hashed))
+
+        makedirs(remove_dir, exist_ok=True)
+
+    io.debug(_('lock dir for {} is {}'.format(item.name, lock_dir)))
+
+    while True:
```
trehn, line 164: should probably wrap this in `with io.job(_("waiting for download [...]` (and the actual download as well of course)

bundlewrap/items/files.py
```diff
@@ -201,13 +272,19 @@ def content(self):
...
         else:
             return sha1(self.content)
 
     @cached_property
     def template(self):
+        if self.attributes['content_type'] == 'download':
+            file_path, remove_dir = download_file(self)
+            if remove_dir:
+                io.debug(_("registering {} for deletion on exit").format(remove_dir))
+                at_exit(rmtree, remove_dir, ignore_errors=True)
```
trehn, line 286: Ugh, this is tricky. Initially I was wondering why you're using your own filesystem locking instead of Python locks. Then I thought "neat, that way other bw processes can share the cache". But with `at_exit` you might then delete stuff those other processes are still using. Off the top of my head there is no reasonable way to fix this and we should make the download cache per-process (and thus don't need the fs locking anymore).

bundlewrap/items/files.py
```diff
@@ -476,10 +563,29 @@ def validate_attributes(cls, bundle, item_id, attributes):
...
+        if attributes.get('content_type') == 'download':
+            if 'source' not in attributes:
+                raise BundleError(_(
+                    "{item} from bundle '{bundle}' is of type 'download', but missing "
+                    "required attribute 'source'"
+                ).format(item=item_id, bundle=bundle.name))
+            elif '://' not in attributes['source']:
+                raise BundleError(_(
+                    "{item} from bundle '{bundle}' is of type 'download', but {source} "
+                    "does not look like an url"
```
trehn, line 581:
```suggestion
                    "does not look like a URL"
```

bundlewrap/items/files.py
trehn: Nevermind.

trehn:
Thanks!

PR #677, opened 2021-10-15
add pkg_pamac to include AUR support for arch distros

simonhoffmann:
Adds a new item `pkg_pamac`, which uses the AUR helper pamac to install packages. Pamac supports both official repos and AUR packages, thus a switch `aur` is introduced, telling pamac to install from repo or build from AUR.


trehn:
Also, this should use `get_auto_deps()` to raise a `BundleError` if a `pkg_pacman` item for the same package exists (similar to how files and symlinks block each other https://github.com/bundlewrap/bundlewrap/blob/main/bundlewrap/items/directories.py#L197)

PR #684, opened 2021-11-02
Do more package items need "block concurrent runs"?

vain:
A couple of items allow multiple runs in parallel. This might work if you're installing multiple *different* packages, but what if those package pull in the *same* dependencies? I can't imagine that that's healthy. 🤔

This PR only changes `pkg_pip`, but there's more. Let's discuss first, though.

### DIFF
--- a/bundlewrap/items/pkg_pip.py
+++ b/bundlewrap/items/pkg_pip.py
@@ -17,6 +17,10 @@ class PipPkg(Item):
     }
     ITEM_TYPE_NAME = "pkg_pip"
 
+    @classmethod
+    def block_concurrent(cls, node_os, node_os_version):
+        return [cls.ITEM_TYPE_NAME]
+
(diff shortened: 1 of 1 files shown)


trehn:
TBH I don't recall ever seeing any package manager handling that by waiting for the other install to finish. So it's probably best to do that for all packages.

fkusei:
The only package manager i know which waits for other runs to finish is `yay`, but we don't have any items for that.

simonhoffmann:
Actually, the class `Pkg` has exactly the three lines as added by @vain.
So I'd like to raise the question, why pkgs freebsd, openbsd and zypper (as well as pip of course), are not a Pkg-Subclass, but an Item-Subclass

trehn:
added it for the others in 5817c66b65a4020ccf5f00b2cacc3f20b02c5455

PR #686, opened 2021-11-12
[poc] [rfc] item zfs_dataset: tidy up previously changed properties

CroneKorkN:
Zfs_dataset properties can orphan easily, because previously changed values are currently not considered by the item, unless explicitly set. We could address that by looking for changed properties via `zfs get -o local`.

### DIFF
--- a/bundlewrap/items/zfs_dataset.py
+++ b/bundlewrap/items/zfs_dataset.py
@@ -3,137 +3,154 @@ from pipes import quote
 from bundlewrap.exceptions import BundleError
 from bundlewrap.items import Item
 from bundlewrap.utils.text import mark_for_translation as _
+from bundlewrap.utils import cached_property
 
 
 class ZFSDataset(Item):
     """
-    Creates ZFS datasets and manages their options.
+    Creates ZFS datasets and manages their properties.
     """
     BUNDLE_ATTRIBUTE_NAME = "zfs_datasets"
     REJECT_UNKNOWN_ATTRIBUTES = False
     ITEM_TYPE_NAME = "zfs_dataset"
+    # for defaults which should be different from 'zfs inherit -S' behaviour
+    PROPERTY_DEFAULTS = {
+        'mountpoint': 'none',
+    }
 
     def __repr__(self):
-        return f"<ZFSDataset name:{self.name} {' '.join(f'{k}:{v}' for k,v in self.attributes.items())}>"
+        return f"<ZFSDataset name:{self.name} {' '.join(f'{k}:{v}' for k,v in self.__item_properties.items())}>"
 
-    def __create(self, path, options):
-        option_list = []
-        for option, value in sorted(options.items()):
-            # We must exclude the 'mounted' property here because it's a
-            # read-only "informational" property.
-            if option != 'mounted' and value is not None:
-                option_list.append("-o {}={}".format(quote(option), quote(value)))
-        option_args = " ".join(option_list)
+    # HELPERS
 
-        self.run(
-            "zfs create {} {}".format(
-                option_args,
-                quote(path),
-            ),
-            may_fail=True,
-        )
+    def __zfs(self, cmd):
+        return self.run(f'zfs {cmd}').stdout.decode('utf-8').strip()
 
-        if options['mounted'] == 'no':
-            self.__set_option(path, 'mounted', 'no')
-
-    def __does_exist(self, path):
-        status_result = self.run(
-            "zfs list {}".format(quote(path)),
-            may_fail=True,
-        )
-        return status_result.return_code == 0
-
-    def __get_option(self, path, option):
-        cmd = "zfs get -Hp -o value {} {}".format(quote(option), quote(path))
-        # We always expect this to succeed since we don't call this function
-        # if we have already established that the dataset does not exist.
-        status_result = self.run(cmd)
-        return status_result.stdout.decode('utf-8').strip()
-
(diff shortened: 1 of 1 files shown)


bundlewrap/items/zfs_dataset.py
```diff
@@ -3,137 +3,154 @@
...
-                    self.attributes.get('mountpoint') and
-                    item.attributes.get('mountpoint') and
-                    self.attributes['mountpoint'].startswith(item.attributes['mountpoint'])
+                # add dependency to parent dataset
+                needs.add(item.id)
+            elif self.__item_properties.get('mountpoint'):
+                parent_directory = '/'.join(self.__item_properties.get('mountpoint', '').split('/')[0:-1])
+                if (
+                    item.ITEM_TYPE_NAME == "zfs_dataset" and
+                    item.attributes.get('mountpoint') == parent_directory or
```
fkusei, line 143: This now won't work if we have datasets which get mounted to `/foo` and `/foo/bar/baz`. The old mechanism did take that into account.

bundlewrap/items/zfs_dataset.py
```diff
@@ -3,137 +3,154 @@
...
-        option_args = " ".join(option_list)
-
-        self.run(
-            "zfs create {} {}".format(
-                option_args,
-                quote(path),
-            ),
-            may_fail=True,
-        )
+        return f"<ZFSDataset name:{self.name} {' '.join(f'{k}:{v}' for k,v in self.__item_properties.items())}>"
```
fkusei, line 22: `__repr__()` should not need to run anything on the node. Calling `__item_properties()` will try to get the properties from the node itself.

bundlewrap/items/zfs_dataset.py
```diff
@@ -3,137 +3,154 @@
...
     """
-    Creates ZFS datasets and manages their options.
+    Creates ZFS datasets and manages their properties.
     """
     BUNDLE_ATTRIBUTE_NAME = "zfs_datasets"
     REJECT_UNKNOWN_ATTRIBUTES = False
     ITEM_TYPE_NAME = "zfs_dataset"
+    # for defaults which should be different from 'zfs inherit -S' behaviour
+    PROPERTY_DEFAULTS = {
+        'mountpoint': 'none',
```
fkusei, line 18: Don't we want to default to `None` here?

bundlewrap/items/zfs_dataset.py
```diff
@@ -3,137 +3,154 @@
...
-                # XXX Could be optimized by finding the "largest"
-                # parent only.
-                if self.name.startswith(item.name + "/"):
-                    needs.add(item.id)
-                elif (
-                    self.attributes.get('mountpoint') and
-                    item.attributes.get('mountpoint') and
-                    self.attributes['mountpoint'].startswith(item.attributes['mountpoint'])
+                # add dependency to parent dataset
+                needs.add(item.id)
```
fkusei, line 138: We should also check wether the parent dataset does exist (like we already do for pool items). Otherwise the creation will fail.

bundlewrap/items/zfs_dataset.py
CroneKorkN: Only depending on the direct parent was a reaction to the comment:
```                # XXX Could be optimized by finding the "largest"
                # parent only.
```
But i see the descriptancy between "largest parent item" and "parent directory".

bundlewrap/items/zfs_dataset.py
CroneKorkN: None would result in the zfs default value, which is mounting a dataset like `foo/bar/baz` to `/foo/bar/baz`.

CroneKorkN:
The details aside, my biggest concern is that i generally only found kind of ugly ways to implement this.

Kunsi:
Revisiting this: Is it important to show the changed properties in the `cdict`? That's the only reason why it has to be that complex, right?

If we don't need to have the properties in the cdict, you could do something like this:
```diff
diff --git a/bundlewrap/items/zfs_dataset.py b/bundlewrap/items/zfs_dataset.py
index e4d8f5be..429d2a60 100644
--- a/bundlewrap/items/zfs_dataset.py
+++ b/bundlewrap/items/zfs_dataset.py
@@ -41,13 +41,6 @@ class ZFSDataset(Item):
         )
         return status_result.return_code == 0
 
-    def __get_option(self, path, option):
-        cmd = "zfs get -Hp -o value {} {}".format(quote(option), quote(path))
-        # We always expect this to succeed since we don't call this function
-        # if we have already established that the dataset does not exist.
-        status_result = self.run(cmd)
-        return status_result.stdout.decode('utf-8').strip()
-
     def __set_option(self, path, option, value):
         if option == 'mounted':
             # 'mounted' is a read-only property that can not be altered by
@@ -120,7 +113,9 @@ class ZFSDataset(Item):
             return None
 
         sdict = {}
-        for option in self.attributes:
-            sdict[option] = self.__get_option(self.name, option)
-        sdict['mounted'] = self.__get_option(self.name, 'mounted')
+        options = self.run(f'zfs get all {quote(self.name)} -H -p -o property,value,source')
+        for line in options.stdout.decode('UTF-8').splitlines():
+            prop, value, source = line.strip().split('\t', 2)
+            if prop in self.attributes or source == 'local' or prop in ('mounted',):
+                sdict[prop] = value
         return sdict
```

vain:
Talked to @CroneKorkN, this can be closed.

PR #694, opened 2021-12-24
items/zfs_dataset: explicitely set mountpoint and canmount, if not already specified

Kunsi:
This fixes an issue on some systems, if you had `tank/myapp` and `tank/myapp/somewhere`, but only mounted the latter dataset. The item wouldn't specify a mountpoint for `tank/myapp`, resulting in mounting the dataset to `/tank/myapp`.

That alone wouldn't be a problem, but bundlewrap then sees 'oh, that dataset is mounted, but shouldn't be, lets unmount it'. Unmounting fails, because `tank/myapp/somewhere` is still mounted (as wanted).

### DIFF
--- a/bundlewrap/items/zfs_dataset.py
+++ b/bundlewrap/items/zfs_dataset.py
@@ -137,3 +137,17 @@ class ZFSDataset(Item):
             sdict[option] = self.__get_option(self.name, option)
         sdict['mounted'] = self.__get_option(self.name, 'mounted')
         return sdict
+
+    def patch_attributes(self, attributes):
+        # If no mountpoint was specified, explicitely set it to 'none'.
+        # Also, set `canmount` to 'off', so we don't accidentially mount
+        # it.
+        if 'canmount' not in attributes:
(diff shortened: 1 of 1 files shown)


trehn:
This breaks inherited mountpoints. The whole idea behind `canmount` is that you can set `canmount=off` *and* a mountpoint for `foo` at `/tank/foo` and then automatically have `foo/bar` mounted at `/tank/foo/bar` *without* having to set that mountpoint explicitly.

Kunsi:
hm, so we'd have to check wether any of the parent datasets do have a mountpoint set before changing those attributes.
Is that possible inside `patch_attributes`?

If not, should we just set the mountpoint option? Do people even expect those inheritances if a mountpoint for the parent dataset was set?

trehn:
There is no guarantee that bw even manages the parent dataset at all, so no.

Mountpoint inheritance is a documented feature, so it's not our place to assume users don't expect it.

Don't think there is anything to be done here, this is a bundle problem.

PR #699, opened 2022-02-06
cache uploads in /tmp, which is more likely writebale than ~

CroneKorkN:


Kunsi:
This should probably be configurable on a per-node level, like we already have for `pip_command` and such

PR #712, opened 2022-06-05
add items/dconf

Kunsi:
Adds the ability to manage all sorts of GNOME related settings.


trehn:
commands used must be added to `docs/content/guide/installation.md`

trehn:
rebase to fix test failure

bundlewrap/items/dconf.py
```diff
@@ -0,0 +1,124 @@
+from bundlewrap.exceptions import BundleError
+from bundlewrap.items import Item
+from json import dumps
+from ast import literal_eval
```
trehn, line 4: stdlib on top plz: https://pep8.org/#imports

bundlewrap/items/dconf.py
```diff
@@ -0,0 +1,126 @@
...
+                f'Item {item_id} in bundle {bundle} has no value set. '
+                'Please explicitely set the "reset" attribute if you '
+                'wish to reset this setting to the default value.'
+            )
+        if not isinstance(attributes.get('value', []), (set, list, str, int)):
+            raise BundleError(
+                f'Item {item_id} in bundle {bundle} uses invalid type '
+                'for its "value" attribute, must be of type str, int, '
+                'list, set.'
+            )
```
trehn, line 118: Exception messages in bw are marked for translation, but I'm not sure if that's worth keeping up.

bundlewrap/items/dconf.py
```diff
@@ -0,0 +1,127 @@
...
+
+        return {
+            'value': self.attributes['value'],
+        }
+
+    def sdict(self):
+        result = self.run(f'dconf read {self.path}', may_fail=True)
+        value = self._parse_result(result)
+
+        if not value:
```
Kunsi, line 85: Need to fix this. That if errorneously catches 0 as well. We only want to catch empty strings.

bundlewrap/items/dconf.py
```diff
@@ -0,0 +1,135 @@
...
+        return attributes
+
+    @classmethod
+    def validate_attributes(cls, bundle, item_id, attributes):
+        if not attributes.get('value') and not attributes.get('reset'):
+            raise BundleError(_(
+                f'Item {item_id} in bundle {bundle} has no value set. '
+                'Please explicitely set the "reset" attribute if you '
+                'wish to reset this setting to the default value.'
+            ))
```
trehn, line 115: You can't combine f-strings with `_()` like that. The values have to be inserted into the string *after* it has been through `_()`, so you need to use `_("").format()`.

trehn: I'll fix after merge

PR #721, opened 2022-11-01
dconf: fix validate_attributes() and _parse_result(), fix docs

fkusei:


bundlewrap/items/dconf.py
```diff
@@ -107,7 +109,7 @@ def patch_attributes(self, attributes):
 
     @classmethod
     def validate_attributes(cls, bundle, item_id, attributes):
-        if not attributes.get('value') and not attributes.get('reset'):
+        if 'value' not in attributes and 'reset' not in attributes:
```
trehn, line 112: setting `reset` to `False` should be handled here, `cdict()` also demands a true-ish value for it to work

trehn: the docs should probably also mention what value `reset` should be set to

PR #723, opened 2022-11-11
Allow removal of essentials with pkg_apt

hexchen:
Currently, there is no way to remove essential packages. This change allows to do so by executing all purges with --allow-remove-essential.


trehn:
Thanks!

PR #726, opened 2022-12-06
routeros: Drop connections on errors

vain:
We *think* we saw this:

-   We bombarded a switch with 64 requests at once.
-   The switch took more than 10 seconds to respond for one of those
    requests. This triggers librouteros's timeout during recv().
-   Data for that request might still arrive later on the socket, so one
    of the next invocations of rawCmd() gets a completely unrelated
    answer.

### DIFF
--- a/bundlewrap/items/routeros.py
+++ b/bundlewrap/items/routeros.py
@@ -9,7 +9,8 @@ from bundlewrap.utils.text import mark_for_translation as _
 from librouteros import connect
 
 
-# very basic connection management, connections are never closed
+# very basic connection management, connections are never closed (unless
+# on errors)
 CONNECTIONS = {}
 CONNECTION_LOCK = Lock()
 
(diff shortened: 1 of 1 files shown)


vain:
100% untested due to the lack of a spare switch.

vain:
Fix confirmed.

I also raised the timeout now. We already saw that 10 seconds is too low.

vain:
How to reproduce/test:

-   Set timeout to something like `0.05`, either in our bw code or in `librouteros/__init__.py`.
-   Run `BW_ITEM_WORKERS=1024 bw verify $switch`
-   Possible results:
    -   WRONG: Lots of timeouts and *bad* items (because responses get garbled up).
    -   CORRECT: Lots of timeouts and *unknown* items.
    -   (WRONG: *Only* unknown or good items, which means that your timeout was too low or too high. It's got to be "just right", so that some requests get a timeout and others don't.)

vain:
DeepSource reports an anti-pattern now, but catching *every* exception is the point of this PR.

PR #727, opened 2023-02-15
routeros fixups

vain:


bundlewrap/items/routeros.py
```diff
@@ -124,6 +129,10 @@ def _connection(self):
     def _run(self, *args):
         with CONNECTION_LOCK:
             try:
+                io.debug(
+                    f'{self.node.name}: running routeros command: ' +
+                    repr(args)
```
vain, line 134: How likely is it that this might accidentally log passwords? 😩

fkusei, line 134: you could use `{args!r}` inside the f-string instead of concat-ing it.

trehn: or `{repr(args)}`

vain: I have no preference here. As I wrapped the line anyway, I just used a concat. But I suspect you don’t care about line length here, so I’ll go with `{repr(args)}` since it’s a bit less obscure than `{args!r}`.

PR #729, opened 2023-04-05
improve RouterOS connection handling

trehn:


bundlewrap/operations.py
```diff
@@ -249,6 +256,44 @@ def run(
...
+        # Connection in unknown state, try to close it and then
+        # drop it.
+        with suppress(Exception):
+            connection.close()
+        with suppress(KeyError):
+            with ROUTEROS_CONNECTIONS_LOCK:
+                del ROUTEROS_CONNECTIONS[hostname]
+        raise RemoteException(str(e)) from e
+
+    run_result = RunResult()
```
Kunsi, line 291: What happens to `run_result`? It never gets used.

bundlewrap/items/routeros.py
```diff
@@ -111,44 +107,13 @@ def patch_attributes(self, attributes):
...
-            self._command_results.append({
-                'command': repr(args),
-                'result': run_result,
-            })
-            return result
+    def run_routeros(self, *command):
+        result = self.node.run_routeros(*command)
+        self._command_results.append({
+            'command': repr(command),
+            'result': result,
```
Kunsi, line 114: Doesn't this need to be an instance of `RunResult()`?

bundlewrap/operations.py
```diff
@@ -249,6 +256,45 @@ def run(
...
+                    timeout=120.0,
+                )
+                lock = Lock()
+                ROUTEROS_CONNECTIONS[hostname] = connection, lock
+            except Exception as e:
+                raise RemoteException(str(e)) from e
+
+    try:
+        io.debug(f'{hostname}: running routeros command: {repr(args)}')
+        with lock:
```
vain, line 279: What about this scenario:

1.  Thread A grabs this lock and runs `rawCmd()`.
2.  Thread B wants to do the same and waits at grabbing the lock.
3.  Thread A runs into an error, let's say it's a timeout (this is the most harmful situation, because the connection is still "working" but data is pending; this is what we originally saw). Thread A thus releases the lock.
4.  Thread B grabs the lock and tries to run `rawCmd()`, but the connection is in an undefined state now.
5.  Thread A only now proceeds to run `connection.close()`, but it's too late now, thread B already did its thing.

Should `with lock` be move to the outermost level?

trehn: Yes, but that's still not enough. Had to decouple both locks even more to avoid a deadlock.

bundlewrap/operations.py
```diff
@@ -249,6 +256,52 @@ def run(
...
+                    timeout=120.0,
+                )
+            except Exception as e:
+                raise RemoteException(str(e)) from e
+            else:
+                conn_state['needs_reconnect'] = False
+
+        try:
+            io.debug(f'{hostname}: running routeros command: {repr(args)}')
+            result = tuple(conn_state['connection'].rawCmd(*args))
```
vain, line 292: Code looks good to me. (In theory, `.close()` failing for some odd reason without actually closing the socket could exhaust the process’s file descriptor limit … Should we at least add an `io.debug()` there? The object we’re dealing with here is of type `librouteros.api.Api`, it’s not the raw socket, so errors in the library *are* a possibility.)

Can’t think of any other scenarios that break anything. So, added after line 292 to simulate errors:
```
            if randint(0, 10) > 8:
                raise Exception("Breaking on purpose")
```
Nothing unexpected showed up. Doesn’t mean *for certain* that there aren’t any errors, of course, but I have a good feeling about this.

I’m for merging this. Maybe add the `io.debug()`, your call.

Issue #415, opened 2019-01-23
Detection of systemd enabled services is slightly off


timbuchwaldt:
Systemd knows a state called "enabled-runtime" -> This unit has been required by another unit.
If said other unit has been uninstalled in the meantime the "enabled-runtime" still holds true -> The unit will not be symlinked for start on the next reboot.

PR #730, opened 2023-04-19
svc_systemd: Fix svc_enabled()

vain:
I don't know where the string "runtime-enabled" comes from. It has been called "enabled-runtime" in their manpage since at least 2013.

"runtime-enabled" worked on Ubuntu 18.04, but that's the only reference to it that I can find. It no longer does on Ubuntu 22.04, though.

### DIFF
--- a/bundlewrap/items/svc_systemd.py
+++ b/bundlewrap/items/svc_systemd.py
@@ -32,7 +32,7 @@ def svc_enabled(node, svcname):
     )
     return (
         result.return_code == 0 and
-        force_text(result.stdout).strip() != "runtime-enabled"
+        force_text(result.stdout).strip() != "enabled-runtime"
     )
 
 


vain:
Correction, it’s called `enabled-runtime` even on 18.04. I saw a daemon not launching on boot after an upgrade from 18.04 to 22.04, so I was assuming 18.04 had the string `runtime-enabled`, but a quick test in a VM shows that that’s not true.

It’s `enabled-runtime` even in 16.04:

```
root@ubuntu1604lts:~# systemctl --runtime enable foo
Created symlink from /run/systemd/system/multi-user.target.wants/foo.service to /etc/systemd/system/foo.service.

root@ubuntu1604lts:~# systemctl is-enabled foo
enabled-runtime
```

`runtime-enabled` was introduced for issue #415, but the author called it `enabled-runtime` there as well. Was this just a typo back then?

I’ll edit my commit message in a second.

vain:
I went ahead and removed `runtime-enabled`. I see no reason for it being there.

Did we ever get feedback from @timbuchwaldt if #415 actually got fixed for him?

PR #732, opened 2023-06-19
add `bw ipmi`

Kunsi:
I'm not sure if this is the best way to implement this feature.

Ideally, i'd want to set ipmi login information on a node itself and have a dedicated ipmi command, so i could run `bw ipmi mynodename 'power status'`, but i'm also not sure if *that* is the right way to implement this.

Also, we'd probably need node-specific `BW_IPMITOOL_ARGS`, because Dell iDRAC requires usage of `-I lanplus`, while others might not work using the lanplus protocol.


vain:
We still have a need for this, right?

Scrolling over the diff, I noticed `with ROUTEROS_CONNECTIONS_LOCK:` for the RouterOS commands. Do we need something similar for IPMI?

fkusei:
I need to think about this some more, but yes, i'd like to have this.

Kunsi:
Screenshot 2025-03-02 at 15 50 11

Still needs docs and maybe the feature to run this on multiple nodes. Also, we need error handling.

docs/content/guide/installation.md
```diff
@@ -70,6 +70,7 @@ While the following list might appear long, even very minimal systems should pro
 * `groupmod`
 * `id`
 * `initctl` (only used with [svc_upstart](../items/svc_upstart.md) items)
+* `ipmitool` (only used for [node.run_ipmitool()](api.md))
```
Kunsi, line 73: That's also wrong here, we're running ipmitool locally.

bundlewrap/cmdline/parser.py
```diff
@@ -360,6 +361,39 @@ def build_parser_bw():
         help=_("show config hash for this item on the given node"),
     )
 
+    # bw ipmi
+    help_ipmi = _("Run 'ipmitool' on the ipmit interface of a specific node")
```
vain, line 365: ipmit

bundlewrap/cmdline/parser.py
```diff
@@ -360,6 +361,39 @@ def build_parser_bw():
...
+        type=str,
+        help=HELP_get_target_nodes,
+    )
+    parser_ipmi.add_argument(
+        'command',
+        metavar=_("COMMAND"),
+        type=str,
+        help=_("command to run"),
+    )
+    bw_ipmi_p_default = int(environ.get("BW_NODE_WORKERS", "1"))
```
vain, line 386: Hm, I would have expected `$BW_NODE_WORKERS` to have one default value, not different ones depending on the command. 🤔

Kunsi: Yeah, stubled over this as well, will create a follow-up PR to fix this (and some other things that came to mind when reading this file).

docs/content/guide/api.md
```diff
@@ -194,6 +194,14 @@ Runs a command on the node. Returns an instance of `bundlewrap.operations.RunRes
 -   `may_fail` If `False`, `bundlewrap.exceptions.RemoteException` will be raised if the command does not return 0.
 
 `bundlewrap.exceptions.TransportException` will be raised if there was a transport error while running the command, e.g. the SSH process died unexpectedly.
+=======
+
+<br>
+
+**`.run_ipmitool(command)`**
+
+Runs a command on an ipmi interface. Returns an instance of
```
vain, line 203: This should explain what a “command” is. That should probably include that the `command` is supposed to be a single string, split at whitespace into tokens.

bundlewrap/cmdline/parser.py
vain: How are you planning to fix this, what do you have in mind?

Kunsi: My plan was to create some more HELP_ variables, and DEFAULT_ variables (like we already have HELP_get_target_nodes) to be used here. (But, as stated, not in this PR)

bundlewrap/group.py
```diff
@@ -61,10 +61,11 @@
     'bundles': COLLECTION_OF_STRINGS,
     'cmd_wrapper_inner': str,
     'cmd_wrapper_outer': str,
-    'lock_dir': str,
     'dummy': bool,
     'file_path': str,
+    'ipmi': dict,
```
vain, line 66: Sorry, I only saw this now: Usage of the new `ipmi` attribute isn’t documented yet, right?

Kunsi:
Just noticed the ipmi attributes don't get merged in with group ipmi attributes, needs investigation.

PR #734, opened 2023-11-08
Optimize run_local() in operations.py

vain:
-   Stop using preexec_fn when launching processes. This has the biggest
    impact. Use process_group=0 instead to get the same result. (For
    some reason that I have not investigated, setting preexec_fn -- even
    if it is just an empty function -- has a massive impact on
    performance.)
-   Get rid of the threads. We use select() anyway, so we can just do
    that in the main thread.

### DIFF
--- a/bundlewrap/operations.py
+++ b/bundlewrap/operations.py
@@ -3,9 +3,10 @@ from datetime import datetime
 from shlex import quote
 from select import select
 from shlex import split
-from subprocess import Popen, PIPE
-from threading import Event, Thread, Lock
-from os import close, environ, pipe, read, setpgrp
+from subprocess import Popen
+from sys import version_info
+from threading import Lock
+from os import close, environ, pipe, read, setpgrp, write
 
 from .exceptions import RemoteException
 from .utils import cached_property
@@ -17,28 +18,6 @@ from librouteros import connect
 
 ROUTEROS_CONNECTIONS = {}
 ROUTEROS_CONNECTIONS_LOCK = Lock()
-
-
-def output_thread_body(line_buffer, read_fd, quit_event, read_until_eof):
-    # see run() for details
-    while True:
-        r, w, x = select([read_fd], [], [], 0.1)
-        if r:
-            chunk = read(read_fd, 1024)
-            if chunk:
-                line_buffer.write(chunk)
-            else:  # EOF
-                return
-        elif quit_event.is_set() and not read_until_eof:
-            # one last chance to read output after the child process
-            # has died
-            while True:
-                r, w, x = select([read_fd], [], [], 0)
-                if r:
-                    line_buffer.write(read(read_fd, 1024))
-                else:
-                    break
-            return
 
 
 def download(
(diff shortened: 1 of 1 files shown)


vain:
Since I’ve now tested this function as best as I could, I realized that there is a tiny window where `SIGINT` can still reach the child: Between `fork()` and `setpgid(0, 0)`.

I only saw this once and it is very hard to reproduce, but it can happen:
```
[2023-11-13T07:03:54.303650] [DEBUG] command with ID 5C8C finished with return code -2
```
```
A negative value -N indicates that the child was terminated by signal N (POSIX only).
```
(`SIGINT` is signal number 2.)

If I understand this correctly …

https://github.com/python/cpython/blob/main/Modules/_posixsubprocess.c#L765

… then there is no way to avoid this. We *could* ignore `SIGINT` in worker threads, but it appears that CPython restores signal handlers *before* it calls `setpgid()`, so there will always be a tiny window of opportunity. (*If* I read CPython’s code correctly.)

This has always been an issue, even before this PR.

It’s probably best if we raised a `RemoteException` on negative return codes. Since this is unrelated to this PR, I’d say we make a new one for that issue.

vain:
Regarding interleaving of stdout/stderr:

If you want to be “as real-time as possible”, the best solution is probably threads. Non-threaded, the next best thing is just reading one byte at a time. I can’t think of anything else to solve this problem *on this level*.

We could try to move the problem one level up, and try to solve it in some kind of “unified line buffer” that respects both stdout and stderr. I’m not exactly sure how, though.

But to be honest, I don’t think it’s worth spending more time on this. It’s only relevant for one very narrow use case: Interactive `bw run` with programs running on the node that spit out *large* amounts of data on both stdout/stderr *nearly at the same time*. (In normal operations like `bw apply`, stdout and stderr are split anyway.)

Since I can’t think of a program that does this on a regular basis (might happen on errors, though), I tried to come up with test cases:

1.)
```
bw run node-1 'for i in $(seq 100); do [ "$(expr $i % 2)" = 0 ] && echo stdout || echo stderr >&2; done'
```
This already looks pretty much the same, threads or not. Note that neither is strictly ordered, even with threads I get a couple of adjacent identical lines.

2.)
```
bw run node-1 'big=$(seq 300 | paste -sd,); for i in $(seq 100); do [ "$(expr $i % 2)" = 0 ] && echo "stdout $big"
|| echo "stderr $big" >&2; done'
```
Same.

3.)
```
bw run node-1 'print_data() { local i; for i in $(seq 20); do echo "$1 $i"; done }; for i in $(seq 5); do [ "$(expr $i % 2)" = 0 ] && print_data stdout || print_data stderr >&2; done'
```
This is actually *better* with the new code. I get lots of out-of-order lines with threads.

* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * *

tl;dr: I think we can leave this as is.

vain:
I have done some tests on how it scales. This is a `bw verify` over 32 nodes.

```
BW_NODE_WORKERS=1   {19m 15s}
BW_NODE_WORKERS=2   {10m 4s}
BW_NODE_WORKERS=4   {5m 4s}
BW_NODE_WORKERS=8   {2m 42s}  {2m 45s}
BW_NODE_WORKERS=16  {1m 36s}  {1m 37s}  {1m 38s}
BW_NODE_WORKERS=32  {1m 36s}  {1m 34s}  {1m 37s}
BW_NODE_WORKERS=64  {1m 34s}
BW_NODE_WORKERS=128 {1m 34s}
```

Notes:

-   Multiple times denote multiple runs.
-   My machine has 4 cores (Intel i7-8665U, hyperthreading disabled).
-   Always 8 item workers.
-   Starting around 16 node workers, my CPU is at its limit according to my `htop` and it becomes the major bottleneck. It cannot get much faster beyond this point.
-   Good thing is, though, 32+ workers aren't *slower*, either.
-   64 and 128 workers: Most of them should be idling because there are only 32 nodes.

I have *not* compared this with the old code. The point of this test is to try to show if the current code of this PR introduces odd/unexpected scaling issues, which – I think – it does not.

vain:
To avoid the `vfork()` issue, we *could* think about rewriting this function to use [`posix_spawn`](https://docs.python.org/3/library/os.html?highlight=posix_spawn#os.posix_spawn) instead. It is similarly fast (or faster?), has a well-defined interface and *should* give us the same performance benefit on older versions of Python.

It’s more low-level, though, and I’m not that familiar with it. No idea how it performs on macOS. It’s also not a drop-in replacement (we’d have to implement `shell=True` ourselves).

I might give this a shot next week to see if there is a performance benefit and how big it is. If it’s not a lot, I’d say we focus on other areas first and maybe come back to `posix_spawn` at some later point.

PR #735, opened 2023-11-08
Integration tests: Always check returncode of run()

vain:
Note that the tests in bw_apply_actions.py were mostly without effect.

Maybe we should discuss moving this to `run()` itself because it is easy to forget to perform this check. It is a useful check, though, as it can reveal all kinds of unexpected failures.

### DIFF
--- a/tests/integration/bw_apply_actions.py
+++ b/tests/integration/bw_apply_actions.py
@@ -22,7 +22,8 @@ def test_action_success(tmpdir):
             },
         },
     )
-    run("bw apply localhost", path=str(tmpdir))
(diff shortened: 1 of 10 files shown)


trehn:
Yes, my brain added an implicit `may_fail=False` here from the non-testing `run()`. But I guess explicit is better than implicit, so we should merge it like this?

vain:
Yeah. Maybe it would help to call the testing version something like `testrun()` instead of `run()` to make it more clear that it’s very different. 🤔

trehn:
Nah, I like the name being short. With all the examples you put in place, the correct usage will probably propagate by copy and paste anyway.

Issue #688, opened 2021-11-27
`pkg_pacman` items do partial match when determining which packages are installed


Kunsi:
```✘ kunsi-t470  pacman  pkg_pacman:iptables  failed
│ installed  yes → no
╵
│
├─ pacman -Q iptables (return code: 0)
│ 
│ iptables-nft 1:1.8.7-1
│
├─ pacman --noconfirm -Rs iptables (return code: 1)
│ 
│ error: target not found: iptables
│
├─ pacman -Q iptables (return code: 0)
│ 
│ iptables-nft 1:1.8.7-1
╵
```

Kunsi:
Maybe that's actually a feature of `pacman`, to show packages that 'provide' the corresponding package?

```
/home/kunsi➤ pacman -Qi iptables
Name            : iptables-nft
Version         : 1:1.8.7-1
Description     : Linux kernel packet control tool (using nft interface)
Architecture    : x86_64
URL             : https://www.netfilter.org/projects/iptables/index.html
Licenses        : GPL2
Groups          : None
Provides        : iptables  arptables  ebtables
Depends On      : libnftnl  libpcap  libnfnetlink  libnetfilter_conntrack  bash  nftables
Optional Deps   : None
Required By     : iproute2  systemd
Optional For    : libvirt
Conflicts With  : iptables  arptables  ebtables
Replaces        : None
Installed Size  : 2,41 MiB
Packager        : Antonio Rojas <arojas@archlinux.org>
Build Date      : 2021-01-20T22:31:09 CET
Install Date    : 2021-06-03T13:49:57 CEST
Install Reason  : Explicitly installed
Install Script  : No
Validated By    : Signature
```

PR #738, opened 2023-11-10
pkg pacman fixes

vain:
@Kunsi Please give this a shot regarding #688.


Kunsi:
Yes, this seems to work correctly now. I no longer get false-positives about iptables being installed.

vain:
Needs a comment in the code. 🙃

PR #740, opened 2023-11-11
Fix block_concurrent() not working if item provides canned actions

### LINKED ISSUE #554: svc_openbsd items have to be daisy-chained
I just provisioned a brand new OpenBSD 6.7 box and some services failed to start, because they had not yet been enabled:

```
/etc/rc.d/relayd: need -f to force start since relayd_flags=NO
```

There were two workers doing a `rcctl set $daemon status on` in parallel and I now *suspect* that this must not run in parallel.

Will investigate later.

vain:
If an item type provides canned actions, item.__deps contained dependencies to those canned actions right from the start. This meant that this construct always returned an empty list ("not item.__deps" was never true):

    list(filter(
        lambda item: not item.__deps and item not in processed_items,
        type_items,
    ))

As a result, concurrency blockers were never inserted for such item types.

Should also fix #554.

### DIFF
--- a/bundlewrap/deps.py
+++ b/bundlewrap/deps.py
@@ -269,9 +269,13 @@ def _inject_concurrency_blockers(items, node_os, node_os_version):
         ))
         processed_items = set()
         for item in type_items:
-            # disregard deps to items of other types
+            # disregard deps to items of other types and to canned
+            # actions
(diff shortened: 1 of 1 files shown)


trehn:
🙈

PR #747, opened 2024-01-31
Fix poll() on GNU

vain:
This was racy and didn't show up immediately.

Bringing "elif" back is important, because we must read all available data before we mark the fd for closing.


vain:
FTR: Got a report that this affected macOS as well.

(“This” being fds getting closed too early with data still pending. Manifested as cut-off files in diffs.)

PR #748, opened 2024-02-01
Introduce `BW_DOWNLOAD_TIMEOUT`

zakx:
Hello again!

Calling requests.request() without a timeout results in the call being able to block forever, if the webserver is unable to be reached or does not send a response. In order to keep backwards compatibility, the default behaviour isn't changed and it will still block for everyone who does not set `BW_DOWNLOAD_TIMEOUT`.

Also mentioning the existance of `BW_FILE_DOWNLOAD_CACHE` in the docs.


trehn:
thanks for the idea, but closing in favor of #751

PR #756, opened 2024-05-01
Avoid most of the flickering in some terminals

vain:
I haven’t really tried a lot of different terminals but this affects VTE and that library is used in quite a lot of stuff.

Before:

https://github.com/bundlewrap/bundlewrap/assets/82671/ef34eae7-909b-4592-a36d-7b86e3038586

After:

https://github.com/bundlewrap/bundlewrap/assets/82671/9431e4ca-fa59-4ce0-bed3-a2433a03a73f

I hope the comments in the code explain everything.

### DIFF
--- a/bundlewrap/utils/ui.py
+++ b/bundlewrap/utils/ui.py
@@ -135,13 +135,15 @@ def page_lines(lines):
             io.stdout(line)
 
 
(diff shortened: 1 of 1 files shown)


trehn:
Thank you!

PR #757, opened 2024-05-13
show skipped items at the end of the apply

### LINKED ISSUE #514: apply: Unavailable Faults should indicate “why” they are unavailable
Current situation:

Random guy runs an apply, doesn’t pay close attention. The program exits cleanly and says: “Success.” But “something” still doesn’t work. Guy complains.

What he didn’t realize: One item (out of a few hundred) was skipped, because a Fault was unavailable. It’s even worse when an item uses several Faults.

Maybe we should provide a way to make it more obvious what’s going on here. Which Fault was unavailable and why?

Kunsi:
This fixes #514 by showing a summary of skipped items at the end for all items that provide a more detailed description of the reason they were skipped.

This was implemented by allowing an item to return a tuple for `details` in `Item.apply()`, of which the first value is the technical reason, the second value is the text message that gets shown to the user. For `FaultUnavailable`, we just show the exception message.

screenshot_2024-05-13_053426

### DIFF
--- a/bundlewrap/cmdline/apply.py
+++ b/bundlewrap/cmdline/apply.py
@@ -46,6 +46,7 @@ def bw_apply(repo, args):
     start_time = datetime.now()
     results = []
     skip_list = SkipList(args['resume_file'])
+    skipped_items = set()
 
     def tasks_available():
         return bool(pending_nodes)
@@ -56,12 +57,13 @@ def bw_apply(repo, args):
             'target': node.apply,
             'task_id': node.name,
             'kwargs': {
-                'autoskip_selector': args['autoskip'],
                 'autoonly_selector': args['autoonly'],
+                'autoskip_selector': args['autoskip'],
                 'force': args['force'],
                 'interactive': args['interactive'],
                 'show_diff': args['show_diff'],
                 'skip_list': skip_list,
+                'skipped_item_callback': handle_skipped_item,
                 'workers': args['item_workers'],
             },
         }
@@ -86,6 +88,9 @@ def bw_apply(repo, args):
             io.stderr(msg)
             errors.append(msg)
 
+    def handle_skipped_item(node, item, reason):
+        skipped_items.add((node, item, reason))
+
     worker_pool = WorkerPool(
         tasks_available,
         next_task,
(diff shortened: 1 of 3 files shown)


fkusei:
Immediate Reactions:
- Why are skipped items now shown above and below the summary?
- "Fault unavailable" during the apply should show the reason as well
- Should "Fault unavailable" be an error and thus show up in `error_summary()` and/or change the exitcode?

vain:
Another question: Why only show skipped items? Failed items can be hard to spot as well, if the output is long enough.

This is a tough one. We have to try to keep/make the output clean and concise, and at the same time informative … Maybe ban some of the output to a log file? 🫤

vain:
Personally, what I do is search the terminal history for `✘` or `»`. Typing those symbols can be challenging, though. 🤦

MaZderMind:
I would also argue that this is a UI problem that should be addressed with some care, before jumping to a conclusion.
For my part, I _usually_ don't care about the `dependecy skipped`, I want to see the original reason in the first place, however sometimes one might wonder why an item is not applied.

To me that sounds like we want varying verbosity levels, for example
- 0: Only show errors and a summary of success and ignored at the end
- 1: Also show every success message
- 2: Also show every skipped message but the `dependecy skipped`
- 3: show all messages

We might even give `dependecy skipped` a different color (grey?) and icon do distinguish them better.

Or we might opt to have dedicated Flags for these kinds of log-lines (alas `bw aplly -eosd` for _e_rror, _o_kay, _s_kipped and _d_ependency skipped)

fkusei:
obsolete, see https://github.com/bundlewrap/bundlewrap/pull/785 (however i'd like to see the reason why faults are unavailable, but we have an issue for that)

vain:
> To me that sounds like we want varying verbosity levels

Yes, we do, but the problem is that you don’t know up-front which level you need. So, new flags are not a good idea, I think.

I keep gravitating towards *always* writing a log file (maybe even use the debug log here) and then shortening the normal output. 🤔

vain:
> obsolete, see #785 (however i'd like to see the reason why faults are unavailable, but we have an issue for that)

@fkusei Should we close this PR then?

Issue #557, opened 2020-10-12
bw metadata vs. Faults


trehn:
can't see faults in `bw metadata`, don't immediately resolve them (add `--resolve-faults`)

option to resolve only resolvable Faults and use dummy strings for others

PR #758, opened 2024-05-21
`bw metadata`: Don’t resolve Faults by default

vain:
This implements https://github.com/bundlewrap/bundlewrap/issues/557.

### DIFF
--- a/bundlewrap/utils/__init__.py
+++ b/bundlewrap/utils/__init__.py
@@ -117,6 +117,15 @@ class Fault:
         self.callback = callback
         self.kwargs = kwargs
 
+    def _repr_first(self):
+        if isinstance(self.id_list, list):
+            return f"<Fault: {self.id_list[0]}>"
+        else:
+            return f"<Fault: {self.id_list}>"
+
+    def _repr_full(self):
+        return f"<Fault: {self.id_list}>"
+
     def _resolve(self):
         if self._available is None:
             try:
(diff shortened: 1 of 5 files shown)


bundlewrap/utils/__init__.py
```diff
@@ -117,6 +117,15 @@ def __init__(self, fault_identifier, callback, **kwargs):
         self.callback = callback
         self.kwargs = kwargs
 
+    def _repr_first(self):
+        if isinstance(self.id_list, list):
+            return f"<Fault: {self.id_list[0]}>"
```
fkusei, line 122: I'm not 100% sold on this. If a secret is formatted into something else, i would like to know that. On the other hand, `id_list` might get very long, and we don't want to clutter the whole screen?

vain: It can get very, see reasoning here: https://github.com/bundlewrap/bundlewrap/pull/758/files#diff-58547683697529b0873aa6b387d16f0fefc4e1430e101f8d54198336b62a7b31R305-R312

The usual `__repr__()` does this:

image

While `_repr_first()` does this:

image

vain: … I just had a use case where the *full list* would have been helpful. Need to think about this.

`--unresolved-faults-full` would do the trick, but feels quite clunky. 🫤

vain: I suggest merging this as is. `--unresolved-faults-full` is probably very rare. Let’s see how often we actually need that.

PR #760, opened 2024-06-20
bw lock: Don't do anything if there are no targets

vain:
This used to crash if there were no targets whatsoever:

    $ bw lock add foo -e 1m -c test
    » foo  is a dummy node
      File "/home/user/git/bundlewrap/bundlewrap/utils/cmdline.py", line 71, in wrapper
        return f(*args, **kwargs)
               ^^^^^^^^^^^^^^^^^^
      File "/home/user/git/bundlewrap/bundlewrap/cmdline/__init__.py", line 77, in main
        pargs.func(repo, text_pargs)
      File "/home/user/git/bundlewrap/bundlewrap/cmdline/lock.py", line 44, in bw_lock_add
        max_node_name_length = max([len(node.name) for node in target_nodes])
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ValueError: max() iterable argument is empty

### DIFF
--- a/bundlewrap/cmdline/lock.py
+++ b/bundlewrap/cmdline/lock.py
@@ -40,6 +40,9 @@ def bw_lock_add(repo, args):
     errors = []
     target_nodes = get_target_nodes(repo, args['targets'], args['node_workers'])
     target_nodes = remove_dummy_nodes(target_nodes)
+    if not target_nodes:
+        return
+
     pending_nodes = target_nodes[:]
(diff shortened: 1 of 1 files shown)


vain:
I thought about doing `max_node_name_length = max([len(node.name) for node in target_nodes] or [''])` for a moment, because that’s the only thing that actually _fails_ here. But why run the rest of the function and launch the worker pools and everything if we know for a fact that nothing is going to happen?

PR #761, opened 2024-06-25
items/routeros: update all attributes of an item at once

MaZderMind:
this fixes issues where one attibute is not valid without the other being set, ie it is not allowed to add a port as untagged to a vlan as long as it is still tagged member of the same vlan. this change also speeds up the overall apply because fewer requests have to be made.


trehn:
Nice, thank you!

PR #764, opened 2024-06-27
items/pkg_apt: mark packages as 'explicitely installed'

### LINKED ISSUE #584: Should bw mark packages as “explicitly installed”?
Scenario:

-   You install package foo.
-   foo has a dependency on bar, so that one gets installed as well.
-   bar is a bw item, too, and is supposed to be installed.
-   You do a system upgrade or something else that removes bar from the list of dependencies from foo.
-   bw sees both packages as installed and is happy.
-   Something like `apt autoremove` might pick up bar as an orphaned package and remove it.

Sounds like an academic example, but we believe this just happened to us.

Should bw mark all packages as “explicitly installed” to prevent tools like `apt autoremove` from removing such packages?

Kunsi:
This fixes #584.

Without marking a package as explicitely installed, a package might come in as a dependency of another package without bundlewrap noticing it. This would allow a mechanism such as `apt autoremove` to remove the package, even though we explicitely want this package to be installed, leading to possible downtime.

For this to work, i had to override some methods imported from the generic `Pkg` class, which seems fine to me.

```
➤ bw items carlene pkg_apt:weechat
{
    "installed": true,
    "mark": "manual"
}
➤ bw items carlene pkg_apt:weechat --state
{
    "installed": true,
    "mark": "manual"
}
```

Has been tested on a couple of systems of my own, and of course i had some packages installed as dependencies.

### DIFF
--- a/bundlewrap/items/pkg_apt.py
+++ b/bundlewrap/items/pkg_apt.py
@@ -1,3 +1,4 @@
+from contextlib import suppress
 from shlex import quote
 
 from bundlewrap.exceptions import BundleError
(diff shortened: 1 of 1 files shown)


trehn:
Is there a reason for this to be a str instead of a bool?

Also needs docs.

trehn:
Actually just had to revert this because it's a performance regression. Now we once again need to run a command for every single package on the node when creating the sdict. The whole point of the pkg_installed_cache was to avoid that. So now we need to fetch all the packages marked as manual in one go and stick them in a similar cache.

PR #766, opened 2024-06-28
Add runlevel support to svc_openrc

robertgzr:
This allows registering services to e.g. the "boot" runlevel instead of
forcing "default"

Signed-off-by: Robert Günzler <r@gnzler.io>


trehn:
Thanks!

PR #775, opened 2024-09-20
Items postgres role support modern algorithms

SeiberTim:
**Problem:**
postgres_role did not accept scram sha 256, because it used CREATE and ALTER, which does not accept anything other than md5 as hash. When passing postgres a password that starts with "md5" it thinks that it is not a password but instead a hash. Anything other that does not start with "md5", postgres trys to hash it.

**Solution:**
Splitted the function so we have one for creating the user and one for setting the password hash directly.

### DIFF
--- a/bundlewrap/items/postgres_roles.py
+++ b/bundlewrap/items/postgres_roles.py
@@ -22,6 +22,10 @@ def fix_role(node, role, attrs, create=False):
         superuser_sql = "SUPERUSER" if attrs['superuser'] else "NOSUPERUSER"
         create_sql = f'CREATE ROLE "{role}" WITH LOGIN {superuser_sql}'
         node.run(f"psql -nqw -c {quote(create_sql)}", user="postgres")
+    else:
+        superuser_sql = "SUPERUSER" if attrs['superuser'] else "NOSUPERUSER"
+        alter_superuser_sql = f'ALTER ROLE "{role}" {superuser_sql}'
+        node.run(f"psql -nqw -c {quote(alter_superuser_sql)}", user="postgres")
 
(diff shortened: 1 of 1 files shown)


bundlewrap/items/postgres_roles.py
```diff
@@ -17,14 +18,13 @@ def delete_role(node, role):
...
-    password = " PASSWORD '{}'".format(attrs['password_hash'])
-    sql = "{operation} ROLE \\\"{role}\\\" WITH LOGIN {superuser}SUPERUSER{password}".format(
-        operation="CREATE" if create else "ALTER",
-        password="" if attrs['password_hash'] is None else password,
-        role=role,
-        superuser="" if attrs['superuser'] is True else "NO",
-    )
-    node.run(f"psql -nqw -c \"{sql}\"", user="postgres")
+    if create:
+        superuser_sql = "SUPERUSER" if attrs['superuser'] else "NOSUPERUSER"
```
vain, line 22: I’m afraid this is a breaking change. The superuser property no longer gets updated in the database.

Suppose there’s an existing user with the item attribute `superuser = True`. When changing this attribute to `False`, it must be changed in the DB as well. That’s not the case currently:

```
? foo-test  ╭─ postgres_role:testpho
? foo-test  │
? foo-test  │  superuser  yes → no
? foo-test  │
? foo-test  ╰─ Fix postgres_role:testpho? [Y/n] y
? foo-test
✘ foo-test  postgresql  postgres_role:testpho  failed
│ superuser  yes → no
╵
```

vain:
Alright, LGTM, works in my tests. 👍

PR #776, opened 2024-09-25
Rework "apt manually installed"

fkusei:
This should fix the concerns stated in https://github.com/bundlewrap/bundlewrap/pull/764 about being a performance regression.

### DIFF
--- a/bundlewrap/items/pkg_apt.py
+++ b/bundlewrap/items/pkg_apt.py
@@ -17,6 +17,12 @@ class AptPkg(Pkg):
     }
     _pkg_manual_cache = {}
 
+    def pkg_all_installed(self):
+        result = self.run("dpkg -l | grep '^ii'")
+        for line in result.stdout.decode('utf-8').strip().split("\n"):
+            pkg_name = line[4:].split()[0].replace(":", "_")
+            yield "{}:{}".format(self.ITEM_TYPE_NAME, pkg_name)
+
     def cdict(self):
         return {
             'installed': self.attributes['installed'],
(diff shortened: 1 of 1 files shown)


fkusei:
There seems to be another regression, because sometimes marking works, but the cache is not updated, possibly i'm doing something wrong here.

fkusei:
Yes, i've held it wrong.

bundlewrap/items/pkg_apt.py
```diff
@@ -14,8 +15,46 @@ class AptPkg(Pkg):
...
+                self.pkg_remove()
+            else:
+                self._pkg_manual_cache.setdefault(self.node.name, set()).add(pkg_name)
+                self.pkg_install()
+        elif 'mark' in status.keys_to_fix:
+            self._pkg_manual_cache.setdefault(self.node.name, set()).add(pkg_name)
+            self.run("apt-mark manual {}".format(quote(pkg_name)))
 
     def pkg_all_installed(self):
+        result = self.run("apt-mark showmanual")
```
vain, line 53: Populating the cache in this function (`pkg_all_installed()`) is racy, sadly. It’s not guaranteed that this function gets called *before* we query the cache for the first time.

Either use the pattern as `pkg_installed_cached()` / `pkg_all_installed()`, or maybe just move populating the cache into `pkg_manually_installed()`?

fkusei: In my tests, this worked every time. I have moved the function as requested nonetheless.

vain:
Alrigthy, thanks! 👍

PR #778, opened 2024-10-23
items/pkg_apt: ensure we only run one `apt-mark showmanual` process at a time

fkusei:


fkusei:
wait, this can be done simpler, i think

bundlewrap/items/pkg_apt.py
```diff
@@ -105,11 +107,15 @@ def validate_attributes(cls, bundle, item_id, attributes):
             ))
 
     def pkg_manually_installed(self):
-        if self.node.name not in self._pkg_manual_cache:
-            result = self.run("apt-mark showmanual")
-            self._pkg_manual_cache[self.node.name] = set()
-            for line in result.stdout.decode('utf-8').strip().splitlines():
-                self._pkg_manual_cache[self.node.name].add(line.strip())
+        if self.node.name not in APT_SHOWMANUAL_LOCK:
```
vain, line 110: This isn’t thread-safe, is it? Am I missing something? 🤔

-   Thread A: Determines that `APT_SHOWMANUAL_LOCK` isn’t in the dictionary.
-   Thread B: Determines that `APT_SHOWMANUAL_LOCK` isn’t in the dictionary.
-   Thread A:
    -   Creates a new `Lock` object and stores it in the dictionary.
    -   Continues to line 113, acquires lock.
-   Thread B:
    -   Only now does thread B execute the body of the `if`.
    -   Creates a new `Lock` object and stores it in the dictionary.
-   Thread A: Continues to do some part of the `with` body.
-   Thread B:
    -   Reaches line 113.
    -   Acquires a *different* `Lock` object than thread A.

→ Both threads execute the `with` body concurrently.

fkusei: I'm not sure. The ideal case would be that `block_concurrent()` would also block concurrent calls to `cdict()`, then we wouldn't have to deal with locks here.

However i don't see any other method of doing a per-node locking of running `apt-mark showmanual`. The only way i can see would be to use a global lock, which feels wrong to me.

MaZderMind: How about a Class managing this:
```
from threading import Lock


class LockDict():
    def __init__(self):
        self._access_lock = Lock()
        self._lock_dicts = dict()

    def lock_for(self, name):
        with self._access_lock:
            if name in self._lock_dicts:
                return self._lock_dicts[name]
            else:
                self._lock_dicts[name] = Lock()
                return self._lock_dicts[name]

```

used like this:
```
d = LockDict()
def do_somethoing():
    with d.lock_for('test'):
        print('test')
```

That would unambiguously show what kind of locking structure we expect.

MaZderMind (changes requested):

fkusei:
Obsolete with https://github.com/bundlewrap/bundlewrap/pull/781

vain:
> Obsolete with #781

Hm, not sure, that PR doesn’t touch `bw verify`. 🤔

fkusei:
You're right.

bundlewrap/items/pkg.py
```diff
@@ -47,12 +67,15 @@ def pkg_installed(self):
         raise NotImplementedError
 
     def pkg_installed_cached(self):
-        cache = self._pkg_install_cache.setdefault(self.node.name, set())
+        # ensure we don't run pkg_all_installed() concurrently, breaks
+        # some package managers
+        with PKG_INSTALLED_LOCK.lock(self.node.name):
```
fkusei, line 72: I decided to add this here, because i believe we have the same issue here.

vain: Yes, we do: https://github.com/bundlewrap/bundlewrap/issues/782 👍

vain:
LGTM and works as intended. ✅

Unless I hear otherwise from @MaZderMind, I’m going to merge this later today. I’d prefer to do any refactoring (like maybe moving `PkgInstalledLock` its own file or something like that) in a follow-up PR.

MaZderMind (approved):
LGTM ✅

fkusei:
I'm not willing to do a major refactoring of the `Pkg` class in a simple "fix this race condition" PR.

MaZderMind:
I extracted the enhancement requests into #791 and therefore approve this PR as-is.

MaZderMind (approved):

PR #783, opened 2024-10-30
directory: Do not use "rm -rf"

vain:
A directory item wants a certain directory to exist. It does not mean that this directory must be *empty*. There is no need for recursion.


MaZderMind:
The issue failing the build has been around a while and is just triggered by a change to the Directory class. I opend a separate Issue for it: #788

MaZderMind (approved):

Issue #786, opened 2024-11-02
pkg_dnf: support for dnf5


stillbeben:
With the release of Fedora 41 dnf has been upgraded to [dnf5](https://fedoraproject.org/wiki/Changes/SwitchToDnf5).
This release includes changes to the API and CLI. While querying is supported by installing the older dnf versions in parallel, "modifying installed software is not recommended due to differences in system state".

The changes to the CLI result in false positive results when querying a packages install state:

```
[2024-11-02T09:58:14.963101] [DEBUG] Non-zero return code (2) running 'dnf -d0 -e0 list installed systemd-networkd' on 'foobar-impa':

Unknown argument "-d0" for command "dnf5". Add "--help" for more information about the arguments.



[2024-11-02T09:58:14.963808] [DEBUG] worker pool verify_foobar-impa delivering result of foobar-impa:systemd-networkd:pkg_dnf:systemd-networkd on worker #0
[2024-11-02T09:58:14.964382] ✘ foobar-impa  systemd-networkd  pkg_dnf:systemd-networkd  
│ 
│ installed  no → yes
```

As the install action also uses "-d0" as command line option, these actions fail as well.

PR #789, opened 2024-11-05
pkg_dnf: adjustments for dnf5 compatibility

stillbeben:
-d and -e are both deprecated[1]
"dnf list --installed" is compatible with both dnf versions

[1] https://dnf.readthedocs.io/en/latest/command_ref.html

#786


vain:
Heyho and thanks for the patch. 👋

I gave this a shot on Fedora 40 and 41. LGTM.

IIUC, Fedora 39 is basically out of support now. I did not test this PR on that version.

PR #790, opened 2024-11-06
pkg_dnf: Skip header when listing all packages

vain:
This lead to a spurious package called "Installed" to be added to the list.


vain:
Maybe @stillbeben could give this a spin. 🙏

stillbeben:
Tested with Fedora 40 and 41. Works as expected. :)

vain:
Alright, thanks!

PR #801, opened 2025-02-20
bw/items add debug message detailing reasons for missing faults

sschiller-seibert:


bundlewrap/items/files.py
```diff
@@ -67,7 +67,8 @@ def content_processor_jinja2(item):
             repo=item.node.repo,
             **item.attributes['context']
         )
-    except FaultUnavailable:
+    except FaultUnavailable as e:
+        io.debug(f"{item.node.name}:{item.id}: Fault Unavailable - {e}")
```
fkusei, line 71: Can we maybe log `{item.node.name} {item.bundle.name} {item.id}: Fault Unavailable - {e}` to make it easier to find out where to find the item?

sschiller-seibert: Good idea, bundle names were added to most debug outputs in the file item.

Kunsi:
Thanks!

Issue #752, opened 2024-02-28
bw lock should warn for preexisting locks


trehn:
Some users assume that locking a node with `bw lock` is guaranteed to give them exclusive access to that node. They might be surprised when locks by other users already existed, allowing those users access as well.

We could display a warning for overlapping locks. We could also introduce a new kind of lock that cannot be overriden through other locks. For 5.0, we might consider a larger redesign to make locking interactions more intuitive/unambiguous.

vain:
A warning is now shown when locks by other people already exist.

PR #803, opened 2025-03-04
Warn on "bw lock add" if already locked

vain:
CC #752.

Not sure about the warning message that comes before the table:

image


fkusei (approved):

vain:
I just realized that this could have funny side effects if the terminal is too small. bw will launch a pager in that case. So if you call `bw lock add` from a script, this might block the script now.

I guess we should enforce to use a dummy pager here? 🤔

Kunsi (approved):

PR #804, opened 2025-03-06
`bw debug`: add history reading and saving

Kunsi:
It was really annoying to create https://github.com/bundlewrap/bundlewrap/pull/732 without history

### DIFF
--- a/bundlewrap/cmdline/debug.py
+++ b/bundlewrap/cmdline/debug.py
@@ -32,6 +32,12 @@ def bw_debug(repo, args):
         history_filename = join(repo.path, '.bw_debug_history')
         if isfile(history_filename):
             readline.read_history_file(history_filename)
+            previous_history_length = readline.get_current_history_length()
+        else:
+            # touch empty history file so we can append to it (maybe, see
+            # comment below)
+            open(history_filename, 'wb').close()
+            previous_history_length = 0
 
         # set up tab completion
         readline.set_completer(Completer(env).complete)
@@ -40,5 +46,18 @@ def bw_debug(repo, args):
         # launch interactive debug session
         interact(banner=banner, local=env)
 
-        # save history
-        readline.write_history_file(history_filename)
+        # So, the thing is, MacOS does not use GNU readline, but instead
+        # provides you with `editline` as a wrapper. This usually works fine,
+        # except it doesn't implement `append_history_file()` at all.
+        # So we have to find out if append_history_file is supported, and if
+        # not we just overwrite the history file with whatever history the last
+        # closed `bw debug` shell had. This sucks, but that's what we have to
+        # work with.
+        if hasattr(readline, 'append_history_file'):
+            new_history_length = readline.get_current_history_length()
+            readline.append_history_file(
+                new_history_length - previous_history_length,
+                history_filename,
+            )
+        else:
+            readline.write_history_file(history_filename)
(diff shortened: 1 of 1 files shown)


bundlewrap/cmdline/debug.py
```diff
@@ -28,6 +28,17 @@ def bw_debug(repo, args):
...
+
+        # set up tab completion
+        readline.set_completer(Completer(env).complete)
+        readline.parse_and_bind("tab: complete")
+
+        # launch interactive debug session
         interact(banner=banner, local=env)
+
+        # save history
+        readline.write_history_file(history_filename)
```
Kunsi, line 44: Not sure if we should call https://docs.python.org/3/library/readline.html#readline.set_history_length before saving history, and if yes what would be a good number of items to keep.

@vain @MaZderMind any opinions?

Kunsi, line 44: There's also https://docs.python.org/3/library/readline.html#readline.append_history_file - but i have not gotten this to work on my laptop, i always get `AttributeError: module 'readline' has no attribute 'append_history_file'. Did you mean: 'read_history_file'?` (python 3.12.2 on MacOS, `readline._READLINE_VERSION` is 1026, whatever that means)

bundlewrap/cmdline/debug.py
```diff
@@ -1,13 +1,13 @@
+import readline
```
Kunsi, line 1: There's also `gnureadline`, but using that will break tab completion. I think this is the best we can do.

Should we add a warning for apple users that only the history of the last closed `bw debug` shell will get saved?

Kunsi: I guess adding a warning for unintended behaviour where the only feasible solution is "buy another computer" is kinda pointless.

bundlewrap/cmdline/debug.py
```diff
@@ -28,6 +28,36 @@ def bw_debug(repo, args):
...
+        # So, the thing is, MacOS does not use GNU readline, but instead
+        # provides you with `editline` as a wrapper. This usually works fine,
+        # except it doesn't implement `append_history_file()` at all.
+        # So we have to find out if append_history_file is supported, and if
+        # not we just overwrite the history file with whatever history the last
+        # closed `bw debug` shell had. This sucks, but that's what we have to
+        # work with.
+        if hasattr(readline, 'append_history_file'):
+            new_history_length = readline.get_current_history_length()
+            readline.append_history_file(
```
vain, line 58: We want to use `append_history_file()` to allow for multiple terminals running `bw debug`, right?

My only gripe with this is that it will make the history file grow without limit. This can get annoyingliny large over time (resulting in slow startup times). I’ve already had this problem with my normal shell. My only hope is that `bw debug` is used so rarely that it won’t be noticeable.

I’ll just drop the following snippet here (just for future reference, not asking you to use this), which would limit the file size but won’t allow for the “append” feature:

```python
histfile = join(expanduser('~'), '.foobar_history')
try:
    readline.read_history_file(histfile)
    readline.set_history_length(50000)
except FileNotFoundError:
    pass

run()

readline.write_history_file(histfile)
```

Kunsi: I hope so, the [examples](https://docs.python.org/3/library/readline.html#example) use it that way 😅

PR #806, opened 2025-03-10
[QS] bundlewrap/items/git_deploy.py: use ssh BatchMode=yes for git to prev…

CroneKorkN:
…ent hanging on password prompt

### DIFF
--- a/bundlewrap/items/git_deploy.py
+++ b/bundlewrap/items/git_deploy.py
@@ -207,6 +207,7 @@ class GitDeploy(Item):
         """
         git_env = environ.copy()
         git_env['GIT_TERMINAL_PROMPT'] = '0'
+        git_env['GIT_SSH_COMMAND'] = 'ssh -o BatchMode=yes'
         cmdline = ["git"] + cmdline
         io.debug(_("running '{}' in {}").format(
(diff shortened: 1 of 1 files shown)


bundlewrap/items/git_deploy.py
```diff
@@ -207,6 +207,7 @@ def run_git(self, cmdline, repo_dir):
         """
         git_env = environ.copy()
         git_env['GIT_TERMINAL_PROMPT'] = '0'
+        git_env['GIT_SSH_COMMAND'] = 'ssh -o BatchMode=yes'
```
fkusei, line 210: Do we maybe want all of this? https://github.com/bundlewrap/bundlewrap/blob/main/bundlewrap/operations.py#L309-L320

CroneKorkN: Explicitly disabling interactive authentication methods doesnt change the outcome, but it makes sense to not even try those methods in the first place.

StrictHostKeyChecking in my opinion should not be touched here. It should only apply to the node currently being applied.

Kunsi (approved):

fkusei:
Thanks!

PR #808, opened 2025-03-21
move defaults for node_workers, item_workers and softlock_expiry to utils/cmdline.py

Kunsi:
as discussed in #732

### DIFF
--- a/bundlewrap/cmdline/parser.py
+++ b/bundlewrap/cmdline/parser.py
@@ -2,7 +2,10 @@ from argparse import ArgumentParser, RawTextHelpFormatter, SUPPRESS
 from os import environ, getcwd
 
 from .. import VERSION_STRING
-from ..utils.cmdline import HELP_get_target_nodes
+from ..utils.cmdline import (DEFAULT_item_workers, DEFAULT_node_workers,
+                             DEFAULT_softlock_expiry, HELP_get_target_nodes,
+                             HELP_item_workers, HELP_node_workers,
+                             HELP_softlock_expiry)
 from ..utils.text import mark_for_translation as _
 from .apply import bw_apply
 from .debug import bw_debug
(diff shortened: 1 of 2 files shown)


bundlewrap/cmdline/parser.py
```diff
@@ -977,15 +964,12 @@ def build_parser_bw():
         action='store_true',
         dest='stdout_table',
         help=_("include command stdout in stats table"),
-    )
-    bw_run_p_default = int(environ.get("BW_NODE_WORKERS", "1"))
```
Kunsi, line 981: To be discussed: Default was 1 here, but 4 everywhere else.

bundlewrap/cmdline/parser.py
```diff
@@ -690,14 +679,12 @@ def build_parser_bw():
         dest='inline',
         help=_("keep lists on a single line (for grep)"),
     )
-    bw_nodes_p_default = int(environ.get("BW_NODE_WORKERS", "4"))
     parser_nodes.add_argument(
         "-p",
         "--parallel-nodes",
-        default=bw_apply_p_default,
```
Kunsi, line 697: oops 🙃

bundlewrap/cmdline/parser.py
```diff
@@ -383,14 +382,12 @@ def build_parser_bw():
         type=str,
         help=_("command to run"),
     )
-    bw_ipmi_p_default = int(environ.get("BW_NODE_WORKERS", "1"))
```
Kunsi, line 386: Same as for `bw run`, default was 1 instead of 4

bundlewrap/cmdline/parser.py
vain: I guess the reason for setting this to 1 was that you’d actually watch the output of `bw run`, like, a human reads it. It’ll be confusing if lots of different outputs are intermingled.

Personally, I never used `bw run` like this and I’m fine with settings this to `4` consistently.

@trehn Opinions on this?

Kunsi: I assume most of the people using bw professionally will have set `BW_NODE_WORKERS` manually anyway, so this default won't apply. Personally i've always set `-p 1` manually if i need to see the output of running commands in order.

vain (approved):

PR #810, opened 2025-03-21
items: add some more info to __repr__()

Kunsi:


Kunsi:
The radical approach would be to get rid of all `__repr__()` in the items themselves and instead define `Item.__repr__()` something like this:

```python
    def __repr__(self):
        return "<{} {} {}{}".format(
            self.__class__.__name__,
            self.name,
            " ".join([
                f"{k}:{v}"
                for k, v in sorted(self.attributes.items()),
                if "password" not in k
            }),
            (
                " " + " ".join([
                    f"{k}:{v}"
                    for k, v in sorted(self.when_creating.items()),
                    if "password" not in k
                }),
            ) if self.when_creating else "",
        )
```

But i'm not sure if that would be a sensible solution

vain:
I think this is fine as it stands. I’m gonna merge, thx 👍

vain (approved):

PR #811, opened 2025-03-24
utils.ui.IOManager(): support being used as a context manager

Kunsi:
This implements #767

### DIFF
--- a/bundlewrap/utils/ui.py
+++ b/bundlewrap/utils/ui.py
@@ -217,6 +217,14 @@ class IOManager:
         self._status_line_present = False
         self._waiting_for_input = False
 
+    def __enter__(self):
+        self.activate()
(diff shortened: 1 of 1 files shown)


MaZderMind (approved):

vain (approved):

bundlewrap/utils/ui.py
```diff
@@ -217,6 +217,14 @@ def __init__(self):
         self._status_line_present = False
         self._waiting_for_input = False
 
+    def __enter__(self):
+        self.activate()
+        # return self, so users could also do `with IOManager() as io:`
+        return self
+
+    def __exit__(self):
```
Kunsi, line 225: https://docs.python.org/3/reference/compound_stmts.html#with says this needs more arguments 🤔

PR #817, opened 2025-04-29
bw lock show: add quiet mode

sschiller-seibert:
This PR adds a new flag to `bw lock show`, allowing to hide empty lines of nodes which don't have any locks.


fkusei (approved):

fkusei:
Thanks!

PR #825, opened 2025-05-26
cmdline/parser,generate_completions: Add prefix for nodes and groups

jpaul-seibert:
For shell completions in scripts or other tools you might need only nodes or groups as completion.
This allows to differentiate between nodes and groups in the generated target file.

### DIFF
--- a/bundlewrap/cmdline/generate_completions.py
+++ b/bundlewrap/cmdline/generate_completions.py
@@ -4,8 +4,8 @@ from os.path import join
 def bw_generate_completions(repo, args):
     compl_file = join(repo.path, '.bw_shell_completion_targets')
 
-    targets = [node.name for node in repo.nodes]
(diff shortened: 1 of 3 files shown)


bundlewrap/cmdline/generate_completions.py
```diff
@@ -4,8 +4,8 @@
 def bw_generate_completions(repo, args):
     compl_file = join(repo.path, '.bw_shell_completion_targets')
 
-    targets = [node.name for node in repo.nodes]
-    targets.extend([group.name for group in repo.groups])
+    targets = [f'node:{node.name}' for node in repo.nodes]
+    targets.extend([f'group:{group.name}' for group in repo.groups])
```
Kunsi, line 8: Bundlewrap should probably also support this selector: https://github.com/bundlewrap/bundlewrap/issues/827

Kunsi (approved):

PR #829, opened 2025-05-28
Roll our own logic to check requirements.txt

### LINKED ISSUE #794: repo.py depends on deprecated pkg-resources
Hi

repo.py still depends on the deprecated pkg_recources

Use of pkg_resources is deprecated in favor of importlib.resources,
importlib.metadata and their backports (importlib_resources, importlib_metadata).
Some useful APIs are also provided by packaging (e.g. requirements and version
parsing). Users should refrain from new usage of pkg_resources and should work
to port to importlib-based solutions.

Python 3.12 in unstable provides both importlib_resources and
importlib_metadata, so no additional dependencies on those packages are needed.

More information available at:

https://setuptools.pypa.io/en/latest/pkg_resources.html

Debian bug: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1083326

It would be nice if bundlewrap could fix this in an upcoming release.

thanks!

-Jonathan

vain:
`pkg_resources` is deprecated, even spits out a warning now. They don't provide a 1:1 replacement for `require()`.

Use `importlib` to query installed packages and `packaging` to parse requirements.txt lines.

Closes #794.


bundlewrap/repo.py
```diff
@@ -513,32 +515,51 @@ def populate_from_path(self, path):
...
+                if req.marker is not None and not req.marker.evaluate():
+                    continue
+
+                for installed in metadata.distributions():
+                    if req.name.lower() == installed.name.lower():
+                        if req.specifier.contains(installed.version):
+                            # We're good.
+                            break
+                        else:
+                            raise MissingRepoDependency(_(
```
fkusei, line 543: In a follow-up PR we should improve this to show all missing dependencies at once, not just the first one that failed, for better UX.

fkusei (approved):

fkusei (approved):

PR #840, opened 2025-07-30
pkg_apt: allow uninstalling held packages

Kunsi:


fkusei:
To be discussed: Do we also want this for *installing* packages?

fkusei:
@vain Maybe the issue is inside our repo, but i'm not sure. What we do is pinning openjdk to a specific version. Since there are multiple openjdk packages, we first try to fiddle the packages correct (using bw), then adjust the pinning file. When updating openjdk to a new major version, there's the issue that for example openjdk21 is being installed, but 17 is being held and we try to uninstall it. This leads to this error message:
```
The following packages will be REMOVED:
  openjdk17*
The following held packages will be changed:
  openjdk17
0 upgraded, 0 newly installed, 1 to remove and 1 not upgraded.
23 not fully installed or removed.
E: Held packages were changed and -y was used without --allow-change-held-packages.
```

PR #843, opened 2025-08-08
lock: Auto-set identity based on Git branch

### LINKED ISSUE #838: Auto-set identity according to Git branch
I have used a wrapper like this for ages now:

```shell
# Create BW_IDENTITY depending on current branch.
if [ -z "$BW_IDENTITY" ]
then
    branch=$(git rev-parse --abbrev-ref HEAD 2>/dev/null)
    if [ $? -eq 0 ]
    then
        export BW_IDENTITY="$USER@$HOSTNAME:$branch"
    fi
fi
```

This should really be done in core bw.

It allows users to run `bw lock` on one branch, switch branches, and then be prohibited to run `bw apply` on the node they just locked. This is the expected behavior, I think, because otherwise you’d overwrite your own changes.

vain:
Closes #838.


MaZderMind (approved):

Kunsi (approved):

vain:
I’ll go ahead and merge this “as is”, since we’re doing this “bw 5 hack week” now.

PR #844, opened 2025-08-19
items: add mariadb_db and mariadb_user

fkusei:
Based on some `action`s we have in an internal repository. Still needs testing, but i currently don't have a test machine running mariadb.

### DIFF
--- a/bundlewrap/items/mariadb_users.py
+++ b/bundlewrap/items/mariadb_users.py
@@ -1,5 +1,5 @@
 from hashlib import sha1
-from re import match as re_match
+from re import search
 from shlex import quote
 
 from bundlewrap.exceptions import BundleError
@@ -8,12 +8,12 @@ from bundlewrap.utils.text import mark_for_translation as _
 
 
 def hash_password(plaintext):
-    m = sha1()
-    m.update(plaintext)
+    first_hash_interation = sha1()
+    first_hash_interation.update(plaintext.encode())
 
-    n = sha1()
-    n.update(m.digest())
-    return '*' + n.hexdigest().upper()
+    second_hash_interation = sha1()
+    second_hash_interation.update(first_hash_interation.digest())
+    return '*' + second_hash_interation.hexdigest().upper()
 
 
 class MariadbUser(Item):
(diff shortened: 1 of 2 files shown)


bundlewrap/items/mariadb_users.py
```diff
@@ -0,0 +1,123 @@
...
+            password_hash = hash_password(self.attributes['password'])
+            self._query(f"SET PASSWORD FOR '{self.name}' = '{password_hash}';")
+
+        if status.must_be_created or 'all_privileges' in status.keys_to_fix:
+            old_grants = self._get_all_privileges()
+            to_be_removed = old_grants - self.attributes['all_privileges']
+            to_be_added = self.attributes['all_privileges'] - old_grants
+
+            for database in to_be_removed:
+                self._query(f"REVOKE ALL PRIVILEGES ON `{database}`.* FROM '{self.name}';")
```
vain, line 81: So, just for the record: We’re ignoring all kinds of SQL injections, because the user who provides this input is a database admin anyway?

vain:
TODO: Needs docs.

vain:
Back in the day, the reason for not including MySQL items was that you couldn’t be sure that you can actually run MySQL commands on the target system. There was only some *Debian* convention. I forgot the details by now; some config file that is always present on Debian-like systems and that allows root to run DB commands.

But an item in bw core cannot / should not depend on such a thing. It would mean that a user-provided bundle would have to exist in order to be able to use `mysql` items. We had decided against that.

Now, how is this handled in current MariaDB? Are there any prerequisites or is root always allowed to run DB commands?

bundlewrap/items/mariadb_users.py
MaZderMind: Hmmm I don't like this. Event DB Admins might pull in data from elsewhere (LDAP?) and might not be aware when someone has a name with quotes in it.

fkusei: FTR: We do the same thing in postgresql: https://github.com/bundlewrap/bundlewrap/blob/main/bundlewrap/items/postgres_roles.py#L22

PR #845, opened 2025-09-01
Add support for numeric UIDs/GIDs for files and directories

### LINKED ISSUE #841: Allow to use numeric UID/GID for files and directories
Hi,

I would like to implement a feature to be able to use numeric UIDs and GIDs for owner and group of files and directories.
Currently, when using a UID instead of a username, the `chown` works because it recognizes the value as ID but then afterwards the verify fails with

```python
files['/tmp/test'] = {
	'source': 'foo/bar',
	'owner': '123456',
	'group': '123456',
}
```

```
$ bw verify test1
✘ test1  test-bundle  directory:/tmp/test
│ 
│ group  UNKNOWN → 123456
│ owner  UNKNOWN → 123456
╵
```

This is because it's always the name of the owner/group that is used for comparison.

There are however cases in my setup, where I need to let files be owned by UIDs which aren't present in `/etc/passwd` for NFS-shared directories and directories mounted into a uidmapped container namespace owned by high 100000+ UIDs on the host.

Therefore I suggest either adding new attributes `owner_id`/`group_id` or checking whether an int or a string has been given for `owner`/`group`. Then, if it's an UID/GID, use `chown +<UID>:+<GID>` and during verify let `stat` also output the UID/GID of the file/directory.

What do you think about that?

xje4:
There are cases where it's necessary to use chown with a UID/GID instead
of a username because the ID of the owner/group isn't present in /etc/passwd.
This could be because the file is inside a NFS-shared directory or
because it's bind-mounted into a (uidmapped) container environment.

Currently, when using a UID instead of a username, the chown works because
it recognizes the value as ID but then afterwards the verify fails to
verify its action because it compares the output of `stat -c '%U'` with
the `owner` attribute of the File item and `stat` outputs `UNKNOWN` for
UIDs/GIDs it can't map back to a username.

To solve this, support for letting files be owned by a UID/GID instead of
a username is added in form of UIDs/GIDs prefixed by a `+` and if so,
use `stat -c '%u'` for validation to check against the UID instead of the
username.
The +-Syntax is directly taken from `chown` (both BSD and GNU) and therefore
the fix-part just works and only the sdict/check function needs to be adjusted.

Fixes #841

### DIFF
--- a/bundlewrap/items/directories.py
+++ b/bundlewrap/items/directories.py
@@ -242,6 +242,8 @@ class Directory(Item):
         return {'needs': deps}
 
     def sdict(self):
+        use_uid = self.attributes['owner'] is not None and self.attributes['owner'].startswith('+')
+        use_gid = self.attributes['group'] is not None and self.attributes['group'].startswith('+')
         path_info = PathInfo(self.node, self.name)
         if not path_info.exists:
             return None
@@ -252,8 +254,8 @@ class Directory(Item):
             return {
                 'type': 'directory' if path_info.is_directory else path_info.stat['type'],
                 'mode': path_info.mode,
-                'owner': path_info.owner,
-                'group': path_info.group,
+                'owner': ('+' + path_info.owner_id) if use_uid else path_info.owner,
+                'group': ('+' + path_info.group_id) if use_gid else path_info.group,
                 'paths_to_purge': paths_to_purge,
             }
 
--- a/bundlewrap/items/files.py
+++ b/bundlewrap/items/files.py
@@ -441,6 +441,8 @@ class File(Item):
         return deps
 
     def sdict(self):
+        use_uid = self.attributes['owner'] is not None and self.attributes['owner'].startswith('+')
+        use_gid = self.attributes['group'] is not None and self.attributes['group'].startswith('+')
         path_info = PathInfo(self.node, self.name)
         if not path_info.exists:
             return None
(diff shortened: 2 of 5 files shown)


bundlewrap/items/directories.py
```diff
@@ -252,8 +254,8 @@ def sdict(self):
             return {
                 'type': 'directory' if path_info.is_directory else path_info.stat['type'],
                 'mode': path_info.mode,
-                'owner': path_info.owner,
-                'group': path_info.group,
+                'owner': ('+' + path_info.owner_id) if use_uid else path_info.owner,
```
Kunsi, line 257: A thought that ran through my head, but is in no means a "please do that" thing:

We could modify `sdict` so it'll always return both owner/group name and uid/gid, then in `cdict` return either name or id, depending on what the user specified.
This would allow users to be able to easily spot whether a id or name will be changed, instead of having to know "ah, it's prefixed with a plus sign, so it's an id".

What do you think, @vain @MaZderMind?

vain: I’d be fine with both, to be honest. Using the plus sign is probably pretty rare, so I’d say it’s within reason to consult the docs in those cases.

(And the plus sign has the slight advantage of being unambiguous. There’s no potential conflict when you specify both `uid` and `owner`.)

But as I said, I’d be fine with both approaches.

fkusei (approved):

Issue #837, opened 2025-07-07
Python exits silently if target_strings argument in bundlewrap/utils/cmdline.py  get_target_nodes() is a string


sgruber-seibert:
If the target_strings argument for the get_target_nodes function in bundlewrap/utils/cmdline.py is a string instead of a list python exits without an exception/stacktrace with the exit code 1.

 
```
def get_target_nodes(repo, target_strings, node_workers=None):
    if not node_workers:
        node_workers = DEFAULT_node_workers
    targets = set()
    for name in target_strings:
        name = name.strip()
        if name.startswith("bundle:"):
            bundle_name = name.split(":", 1)[1]
            for node in repo.nodes:
                if bundle_name in names(node.bundles):
                    targets.add(node)
        elif name.startswith("!bundle:"):
            bundle_name = name.split(":", 1)[1]
            for node in repo.nodes:
                if bundle_name not in names(node.bundles):
                    targets.add(node)
        elif name.startswith("!group:"):
            group_name = name.split(":", 1)[1]
            for node in repo.nodes:
                if group_name not in names(node.groups):
                    targets.add(node)
        elif name.startswith("lambda:"):
            for node_name, result in _parallel_node_eval(
                repo.nodes,
                name.split(":", 1)[1],
                node_workers,
            ).items():
                if result:
                    targets.add(repo.get_node(node_name))
        else:
            try:
                targets.add(repo.get_node(name))
            except NoSuchNode:
                try:
                    group = repo.get_group(name)
                except NoSuchGroup:
                    io.stderr(_("{x} No such node or group: {name}").format(
                        x=red("!!!"),
                        name=name,
                    ))
                    exit(1)  # Maybe raise an exception instead of exiting silently with exit(1)
                else:
                    targets.update(group.nodes)
    return targets
```

sgruber-seibert:
This occurs when I import the function into a Python script and use a string instead of a list for target_strings there.

vain:
This function isn’t part of [the public API](https://docs.bundlewrap.org/guide/api/), so it’s a bit problematic to use it. If we were to change this function at some point in the future, we would never consider any potential downstream users/breakage.

I realize that there are some functions (like this one) which might be useful in some “general BundleWrap utils” package. Maybe we should collect them and move them to such a package? It might be hard to decide which functions to include, though. And once they *are* part of the public API, we are committed to supporting them. I’m not sure. What do others think about this, @Kunsi @MaZderMind @trehn?

[...]

PR #846, opened 2025-09-08
utils/cmdline: make get_target_nodes raise an exception 

MaZderMind:
…iling silently so it becomes usable from external scripts, provide and use get_target_nodes_or_exit everyhere internal

### DIFF
--- a/bundlewrap/utils/cmdline.py
+++ b/bundlewrap/utils/cmdline.py
@@ -4,7 +4,7 @@ from sys import exit, stderr, stdout
 from traceback import format_exc, print_exc
 
 from ..concurrency import WorkerPool
-from ..exceptions import NoSuchGroup, NoSuchItem, NoSuchNode, RepositoryError
+from ..exceptions import NoSuchGroup, NoSuchItem, NoSuchNode, RepositoryError, UsageException
 from . import names
 from .text import bold, mark_for_translation as _, prefix_lines, red
 from .ui import io, QUIT_EVENT
@@ -198,6 +198,17 @@ def _parallel_node_eval(
     return dict(worker_pool.run())
 
 
+def get_target_nodes_or_exit(repo, target_strings, node_workers=None):
+    try:
+        return get_target_nodes(repo, target_strings, node_workers)
+    except UsageException as e:
+        io.stderr(_("{x} {message}").format(
+            x=red("!!!"),
+            message=str(e),
+        ))
+        exit(1)
+
+
 def get_target_nodes(repo, target_strings, node_workers=None):
     if not node_workers:
         node_workers = DEFAULT_node_workers
@@ -235,11 +246,7 @@ def get_target_nodes(repo, target_strings, node_workers=None):
                 try:
                     group = repo.get_group(name)
                 except NoSuchGroup:
-                    io.stderr(_("{x} No such node or group: {name}").format(
-                        x=red("!!!"),
-                        name=name,
-                    ))
-                    exit(1)
+                    raise UsageException('No such node or group: {}'.format(name))
                 else:
                     targets.update(group.nodes)
     return targets
--- a/bundlewrap/cmdline/apply.py
+++ b/bundlewrap/cmdline/apply.py
@@ -4,7 +4,7 @@ from sys import exit
 from ..concurrency import WorkerPool
 from ..exceptions import GracefulApplyException
 from ..utils import SkipList
-from ..utils.cmdline import count_items, get_target_nodes
+from ..utils.cmdline import count_items, get_target_nodes_or_exit
 from ..utils.table import ROW_SEPARATOR, render_table
 from ..utils.text import (
     blue,
@@ -24,7 +24,7 @@ from ..utils.ui import io
 
 def bw_apply(repo, args):
     errors = []
-    target_nodes = get_target_nodes(repo, args['targets'], args['node_workers'])
+    target_nodes = get_target_nodes_or_exit(repo, args['targets'], args['node_workers'])
     pending_nodes = target_nodes.copy()
 
     try:
--- a/bundlewrap/cmdline/diff.py
+++ b/bundlewrap/cmdline/diff.py
@@ -5,7 +5,7 @@ from sys import exit
 from ..exceptions import NoSuchItem
 from ..metadata import metadata_to_json
 from ..repo import Repository
-from ..utils.cmdline import get_target_nodes
+from ..utils.cmdline import get_target_nodes_or_exit
 from ..utils.dicts import diff_dict, dict_to_text
 from ..utils.scm import get_git_branch, get_git_rev, set_git_rev
 from ..utils.text import (
@@ -322,7 +322,7 @@ def bw_diff(repo, args):
         ).format(x=red("!!!")))
         exit(1)
 
-    target_nodes = sorted(get_target_nodes(repo, args['targets']))
+    target_nodes = sorted(get_target_nodes_or_exit(repo, args['targets']))
 
     if args['branch'] or args['cmd_change'] or args['cmd_reset'] or args['prompt']:
         intermissions = []
--- a/bundlewrap/cmdline/ipmi.py
+++ b/bundlewrap/cmdline/ipmi.py
@@ -1,5 +1,5 @@
 from ..concurrency import WorkerPool
-from ..utils.cmdline import get_target_nodes
+from ..utils.cmdline import get_target_nodes_or_exit
 from ..utils.table import ROW_SEPARATOR, render_table
 from ..utils.text import blue, bold, error_summary, format_duration, green
 from ..utils.text import mark_for_translation as _
@@ -9,7 +9,7 @@ from ..utils.ui import io
 
 def bw_ipmi(repo, args):
     errors = []
-    target_nodes = get_target_nodes(repo, args['targets'], args['node_workers'])
+    target_nodes = get_target_nodes_or_exit(repo, args['targets'], args['node_workers'])
     pending_nodes = target_nodes.copy()
     io.progress_set_total(len(pending_nodes))
 
--- a/bundlewrap/cmdline/lock.py
+++ b/bundlewrap/cmdline/lock.py
@@ -7,7 +7,7 @@ from ..lock import (
     softlock_remove,
     softlocks_to_table,
 )
-from ..utils.cmdline import get_target_nodes
+from ..utils.cmdline import get_target_nodes_or_exit
 from ..utils.text import (
     bold,
     error_summary,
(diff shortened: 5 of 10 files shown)


fkusei (approved):

bundlewrap/utils/cmdline.py
```diff
@@ -235,11 +246,7 @@ def get_target_nodes(repo, target_strings, node_workers=None):
                 try:
                     group = repo.get_group(name)
                 except NoSuchGroup:
-                    io.stderr(_("{x} No such node or group: {name}").format(
-                        x=red("!!!"),
-                        name=name,
-                    ))
-                    exit(1)
+                    raise UsageException('No such node or group: {}'.format(name))
```
fkusei: This was a `mark_for_translation` before 🤔

vain:
Regarding the change itself: It’s a private undocumented API, so I don’t consider this a breaking change.

I’m strongly in favor of not “silently” promoting functions to public stable APIs. Anything that isn’t listed on https://docs.bundlewrap.org/guide/api/ is private. If people use internal functions (just because they *can*, because this is Python), they don’t get to complain when things break.

Honestly, they don’t even get to complain when things don’t work the way they’d like – `get_target_nodes()` is an internal private function and it lives in a file called `cmdline.py`, so it *rightfully* calls `exit()`.

I’m not so sure about this PR, to be honest. We now make a change to accomodate users who *call private functions*. bw itself does not benefit from this change (it, in fact, contributes to spaghettification), only external callers do. We can probably let it slide if it’s just this one function. But please let’s not start treating every *internal* function of bw like it could potentially be called by external users. If we do that, suddenly the whole thing is “public API” and every commit is a breaking change. And I very much doubt that it makes the code better or even easier to read.

If there is an increasing need to extend the public API, then let’s extend the public API. But let’s not pretend that private functions are “kind of public”.

> As the python checker brings it up: Do we prefer f-strings even if the surrounding code still uses .format? Should we refactor the code we touch towards f-strings?

FTR, as discussed elsewhere: The current reason for `.format()` is the usage of `_()`, i.e. “mark for translation”. That won’t work with f-strings.

fkusei:
> If there is an increasing need to extend the public API, then let’s extend the public API. But let’s not pretend that private functions are “kind of public”.

Maybe it would be smart to move `get_target_nodes()` into `bundlewrap.utils`, and have `bundlewrap.utils.cmdline.get_target_nodes()` just call that?

bundlewrap/repo.py
```diff
@@ -562,6 +565,83 @@ def nodes_in_group(self, group_name):
...
+        ).items()
+
+        return [
+            self.get_node(node_name)
+            for node_name, result in result_items
+            if result
+        ]
+
+
+    def nodes_matching(self, target_strings, node_workers=None):
```
fkusei: Is `None` a valid default here? `WorkerPool` expects an integer.

bundlewrap/repo.py
```diff
@@ -562,6 +565,83 @@ def nodes_in_group(self, group_name):
...
+                lambda_str = name.split(":", 1)[1]
+                targets.update(self.nodes_matching_lambda(lambda_str, node_workers))
+            else:
+                try:
+                    targets.add(self.get_node(name))
+                except NoSuchNode:
+                    try:
+                        group = self.get_group(name)
+                    except NoSuchGroup:
+                        raise UsageException(
```
fkusei: Might make sense to return `[]` here and only raise the exception in `bundlewrap.utils.cmdline.get_target_nodes()`. Otherwise we'd have some functions for matching returning empty lists, others raise an Exception.

I think people would expect `repo.nodes_matching()` to work the same as `repo.nodes_in_group()`.

bundlewrap/repo.py
```diff
@@ -562,6 +565,83 @@ def nodes_in_group(self, group_name):
...
+                            _("No such node or group: {name}").format(name=name)
+                        )
+                    else:
+                        targets.update(group.nodes)
+        return targets
```
fkusei: All other matching functions return a list, so we should do the same here (and make a note so we change *all* of them to return sets).

bundlewrap/repo.py
fkusei: Can we be user-friendly and fix #837 as well? (aka, just convert `target_strings` to a list if it's a string)

bundlewrap/repo.py
```diff
@@ -562,6 +565,83 @@ def nodes_in_group(self, group_name):
...
+        """
+        Returns a list of nodes not in the given group.
+        """
+        return [
+            node
+            for node in self.nodes
+            if bundle_name not in names(node.bundles)
+        ]
+
+    def nodes_matching_lambda(self, lamda_str, node_workers=None):
```
fkusei: See comment below about the default value.

bundlewrap/repo.py
MaZderMind: Please note that `nodes_in_any_group` does actually return a generator – but yes, we can settle on list here..

bundlewrap/repo.py
MaZderMind: Yes we can!

bundlewrap/repo.py
MaZderMind: I solved it by allowing `None` for `worker=` on `WorkerPool` which is superior to having the default-value be used all over the place.

bundlewrap/concurrency.py
```diff
@@ -8,6 +8,7 @@
 from .utils.ui import io, QUIT_EVENT
 
 JOIN_TIMEOUT = 5  # seconds
+DEFAULT_WORKERS = 4
```
fkusei, line 11: To discuss: should we use `bundlewrap.utils.cmdline.DEFAULT_node_workers` for that?

MaZderMind: I don't agree, `DEFAULT_node_workers` is coming from the outside, environment or command-line arg and is `cmdline` specific. This default is very intimately tied to the `WorkerPool` which is probably used in a lot of places for different tasks. Usually we expect the number of workers to be passed in from the exterior anyway and this default is more like a last-resort fallback if it is missing.

MaZderMind:
@fkusei I finished the remaining cleanup and added more documentation, also to the public api. Would you mind having another look? Also cc'ing @vain so he can have a look tomorrow.

bundlewrap/concurrency.py
Kunsi: Yeah, makes sense.

bundlewrap/repo.py
```diff
@@ -559,9 +567,124 @@ def nodes_in_any_group(self, group_names):
...
+        ]
+
+    def nodes_matching_lambda(self, lamda_str, lamda_workers=None):
+        """
+        Returns a list of nodes matching the lambda.
+
+        Example:
+            nodes = repo.nodes_matching_lambda("lambda:node.metadata_get('foo/magic', 47) < 3")
+
+        :param lamda_str: string to evaluate as python code with `node` being one of the nodes, expected to return a boolean
```
Kunsi: ```:param lamda_str: string to evaluate as python code with `node` being one of the nodes, result will be evaluated as a boolean```

Returning an integer or a string is totally fine, it will just be cast to boolean and then evaluated (aka 0 is False, 1 and above is True, etc.)

docs/content/guide/api.md
```diff
@@ -94,6 +94,61 @@ Returns a list of Node objects in the named group.
 
 <br>
 
+**`.nodes_not_in_group(group_name)`**
+
+Returns a list of nodes not in the named group.
```
Kunsi: `s/nodes/Node objects/` here and below, to be consistent with what we already have.

docs/content/guide/api.md
```diff
@@ -94,6 +94,61 @@ Returns a list of Node objects in the named group.
...
+<br>
+
+**`.nodes_matching(self, target_strings, lamda_workers=None)`**
+
+Returns a list of nodes matching any of the given target-strings. This is the same API that is used by
+all the bw commandlines, i.e. `bw items` or `bw apply` to select which nodes to operate on.
+
+Example:
+```python
+nodes = repo.nodes_matching_lambda("lambda:node.metadata_get('foo/magic', 47) < 3")
```
Kunsi: This needs to be an example of `.nodes_matching()` instead of `.nodes_matching_lambda()`, if we want to provide an example.

bundlewrap/repo.py
```diff
@@ -559,9 +567,124 @@ def nodes_in_any_group(self, group_names):
...
+            for node in self.nodes
+            if not node.in_group(group_name)
+        ]
+
+    def nodes_with_bundle(self, bundle_name):
+        """
+        Returns a list of nodes that do have the given bundle.
+
+        :param bundle_name: name of the bundle to check for
+        :return list of nodes which do not have the given bundle
```
vain: `:return list of nodes which have the given bundle`

bundlewrap/repo.py
```diff
@@ -559,9 +567,124 @@ def nodes_in_any_group(self, group_names):
...
+        Example:
+            nodes = repo.nodes_matching_lambda("lambda:node.metadata_get('foo/magic', 47) < 3")
+
+        :param lamda_str: string to evaluate as python code with `node` being one of the nodes, expected to return a boolean
+        :param lamda_workers: number of parallel workers used to check lamda condition on every node
+        :return list of nodes matching the given lamda
+        """
+        result_items = parallel_node_eval(
+            self.nodes,
+            lamda_str,
```
vain: `lambda`, in a couple of places.

docs/content/guide/api.md
```diff
@@ -94,6 +94,61 @@ Returns a list of Node objects in the named group.
...
+
+<br>
+
+**`.nodes_without_bundle(bundle_name)`**
+
+Returns a list of nodes that do not have the named bundle.
+
+<br>
+
+**`.nodes_matching_lambda(self, lamda_str, lamda_workers=None)`**
```
vain: Do we need to make this function public? Isn’t it the same as `nodes_matching('lambda:…')`? 🤔

bundlewrap/repo.py
MaZderMind: -> `expected to return a value that can be interpreted as boolean`

docs/content/guide/api.md
MaZderMind: Yes, in the same way as `nodes_in_group(x)` sis the same as `matching('group:x')`. I'd prefer to have separate functions for different logical tasks that can be called without prepending some magic string.

docs/content/guide/api.md
```diff
@@ -94,6 +94,63 @@ Returns a list of Node objects in the named group.
...
+
+Returns a list of Node objects matching the lambda.
+
+Example:
+```python
+nodes = repo.nodes_matching_lambda("node.metadata.get('foo/magic', 47) < 3")
+```
+
+- `lamda_str` is evaluated as python code with `node` being one of the nodes and expected to return a boolean.
```
vain: It still says `lamda` in many places.

MaZderMind: I'm so sorry I missed that. fixed now.

Issue #850, opened 2025-09-21
Add "magic strings" to bundlwrap core?


Kunsi:
In all the repos i use we introduced the concept of "magic strings" to be able to use stuff like `vault.decrypt()` and others.

How do we feel about introducing this concept into bw-core, @trehn @vain @MaZderMind?

I have some ideas in mind which would work nicely i think - i can explain some more at another point in time, or i'll just create a PR.

vain:
Yes. TOML nodes—towards which we’re all gravitating, for some reason—make some sort of “magic string” a necessity, I’m afraid. ☹️

Maybe we can find some “lightweight” implementation and *not* bury this deep inside bw’s core.

trehn:
There has always been a pull in both directions - sometimes we want (#848) or need (`atomic`, `Fault`) proper smart objects in metadata, but we also really want a text representation, both for display with `bw metadata` and for input from TOML files. It would probably be wise to commit to one of these directions.

I'm leaning towards limiting metadata to simple types: bool, int, str, None, list, dict. As for magic strings, I wonder if we can make it as generic and Python-heavy as we tend to do elsewhere. So I think I'd rather not have `"!#__decrypt__(foo)"` and such, but a very generic ```"`foo()`"``` to evaluate Python expressions.

Now what happens if I set ```"`set(1, 2, 3)`"``` on a node? Is the set merged at any point with other metadata? Or does it remain a string until the very end of metadata assembly (meaning merging node- and group-level metadata, reactors, defaults)? The latter is probably much less of a headache.

fkusei:
tbh my thought was "add a config file so users can define their own magic strings" and just evaluate them when loading the toml file - this would be fully compatible with all the existing python code, and consistent with all the "magic string" implementations i've seen so far.

I'd propose against doing something like ```"`...`"```, because in my experience TOML nodes and groups are used to make editing them more user-friendly especially for non-techy users, so in my opinion magic strings like `!decrypt:...` are much easier to understand than ```"`vault.decrypt('....')`"```.

With knowing what magic strings to expect we can also inform the user is they use a magic string we don't already know in a user-friendly way (instead of erroring out with a pythin stacktrace), and - more important - we solve the issue of third-party secrets implementations like `bwtv` or `bwkeepass`. The latter need to be available when evaluating the magic strings, which is somewhere in the bundlewrap core, so we need to import them before calling.

MaZderMind:
[...]

PR #857, opened 2025-10-02
repo: add support for magic strings

Kunsi:
As discussed in #850


Kunsi:
ughhhh, do we need to handle atomic in here as well? 🥴

docs/content/guide/toml.md
```diff
@@ -82,3 +82,29 @@ This should make it pretty straightforward to make changes to lots of nodes with
...
+```python
+@magic_string
+def decrypt(string):
+    return vault.decrypt(string)
+```
+
+In your node you then use the function as follows:
+
+```toml
+mysecret = "!decrypt:SecretStringEncryptedBy `bw pw -e`"
```
vain: It might be more clear if we put it like this:

> In your node you then use the function as follows, where `encrypt$gAAAAABo90x3H...` is the result of `bw pw -e "foo"`:
> 
> ```
> mysecret = "!decrypt:encrypt$gAAAAABo90x3H..."
> ```

This makes it clear that neither the space, nor the backticks, nor the string `SecretStringEncryptedBy` is part of the syntax.

tests/unit/magic_strings.py
```diff
@@ -0,0 +1,95 @@
+from pytest import raises
```
vain, line 1: I’d love to see this file become an integration test, so that we can also cover the `@magic_string` mechanism. Can be a follow-up PR, though. 👍

docs/content/guide/toml.md
```diff
@@ -82,3 +82,29 @@ This should make it pretty straightforward to make changes to lots of nodes with
...
+```
+
+The part between `!` and `:` is used as the function name, everything after the `:` will be passed
+as argument to the called function. Bundlewrap will raise `InvalidMagicStringException` if the
+function cannot be found.
```
vain, line 113: Is there a recommended way to use this new mechanism to declare stuff as `atomic()` in TOML nodes? If so, should we include this in the docs?

This is what I came up with now:

```python
from bundlewrap.metadata import atomic as _bw_atomic


@magic_string
def atomic(arg):
    return _bw_atomic(eval(arg))
```
The `eval()` isn’t particularly pretty, though. 🤔

docs/content/guide/toml.md
Kunsi: Yeah, you're right, providing an example is better than a dummy value.

docs/content/guide/toml.md
Kunsi: I have pushed a more safe example by using `ast.literal_eval()`

tests/unit/magic_strings.py
Kunsi: Yeah, that came to mind as well, but somehow currently i don't feel smart enough to do that. Will look into that later though.

docs/content/guide/toml.md
```diff
@@ -82,3 +82,57 @@ This should make it pretty straightforward to make changes to lots of nodes with
...
+from ast import literal_eval
+from bundlewrap.metadata import atomic as _bw_atomic
+
+@magic_string
+def atomic(arg):
+    # Using `literal_eval()` here will prevent you from most accidents
+    # by only allowing strings, bytes, numbers, tuples, lists, dicts,
+    # sets, booleans, `None` and `Ellipsis`
+    # https://docs.python.org/3/library/ast.html#ast.literal_eval
+    return atomic(literal_eval(arg))
```
vain: That should be `return _bw_atomic(literal_eval(arg))`. 😅

Kunsi: 😵‍💫🥴

vain:
LGTM and marking as ✅, just one last question:

Why call it `magic-strings.py` instead of `magic_strings.py`? The latter would be more “normal”, because you can’t do `import magic-strings` – or is that the point?

vain (approved):

vain (approved):

PR #874, opened 2025-11-27
`bw apply` and `bw verify`: warn users if the `--skip` selector does not match any items

### LINKED ISSUE #873: apply, possibly others: Warn on non-existing tags
I just typed this: `bw apply -s tag:caues-downtime -- mynode` (note the misspelled `tag:causes-downtime`).

What happened was technically correct: bundlewrap skipped all items that matched `tag:caues-downtime`, of which there were zero. This lead to some items with `tag:causes-downtime` to get triggered, causing some downtime.

Bundlewrap should warn users (or maybe even error out on non-interactive applies?) if a tag was specified that matches zero items. This would prevent such issues.

(Possibly needs `--allow-empty-tag` to restore the old behaviour, i'm sure someone somewhere needs this)

"Warning mode" can probably be implemented in 4.x, with "Error mode" in 5.x

Kunsi:
This fixes #873 

... is it actually that simple? Did i miss anything? Will this (significantly) increase startup time?

### DIFF
--- a/bundlewrap/cmdline/verify.py
+++ b/bundlewrap/cmdline/verify.py
@@ -2,7 +2,7 @@ from datetime import datetime
 from sys import exit
 
 from ..concurrency import WorkerPool
-from ..utils.cmdline import count_items, get_target_nodes, verify_autoskip_selector
+from ..utils.cmdline import count_items, get_target_nodes, verify_autoskip_selectors
 from ..utils.table import ROW_SEPARATOR, render_table
 from ..utils.text import (
     blue,
@@ -117,7 +117,7 @@ def bw_verify(repo, args):
     start_time = datetime.now()
     io.progress_set_total(count_items(pending_nodes))
 
-    selectors_not_matching = verify_autoskip_selector(pending_nodes, args['autoskip'])
+    selectors_not_matching = verify_autoskip_selectors(pending_nodes, args['autoskip'])
     if selectors_not_matching:
         io.stderr(_("{x} the following selectors for --skip do not match any items: {selectors}").format(
             x=red("!!!"),
--- a/bundlewrap/utils/cmdline.py
+++ b/bundlewrap/utils/cmdline.py
@@ -161,30 +161,29 @@ def get_target_nodes(repo, target_strings, node_workers=None):
     return nodes_matching
 
 
-
-def verify_autoskip_selector(nodes, autoskip_selector):
-    if not autoskip_selector:
+def verify_autoskip_selectors(nodes, selectors_given):
+    if not selectors_given:
         return set()
 
-    selectors = {
+    selectors_used = {
         selector: False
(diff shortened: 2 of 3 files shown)


Kunsi:
```
/Users/kunsi/git/bundlewrap|main✓ [0]➤ bw verify -s bundle:unbound tag:foobar --exit-on-skiplist-error home.router
!!! the following selectors for --skip do not match any items: tag:foobar

/Users/kunsi/git/bundlewrap|main✓ [0]➤ bw verify -s bundle:nbound tag:foobar --exit-on-skiplist-error home.router
!!! the following selectors for --skip do not match any items: bundle:nbound tag:foobar

/Users/kunsi/git/bundlewrap|main✓ [0]➤ bw verify -s bundle:nbound tag:foobar -- home.router
!!! the following selectors for --skip do not match any items: bundle:nbound tag:foobar
i ╭─────────────┬───────┬──────┬─────┬─────────┬────────┬──────────╮
i │ node        │ items │ good │ bad │ unknown │ health │ duration │
i ├─────────────┼───────┼──────┼─────┼─────────┼────────┼──────────┤
i │ home.router │   305 │  305 │   0 │       0 │ 100.0% │      13s │
i ╰─────────────┴───────┴──────┴─────┴─────────┴────────┴──────────╯
```

bundlewrap/cmdline/verify.py
```diff
@@ -117,6 +117,14 @@ def bw_verify(repo, args):
     start_time = datetime.now()
     io.progress_set_total(count_items(pending_nodes))
 
+    skiplist_errors = verify_autoskip_selector(pending_nodes, args['autoskip'])
```
MaZderMind: `skiplist_errors` seems to be a little too generic, we know that it will contain selectors that did not match (this is also reflected in the error-output below), so I would suggest a specific name: `selectors_not_matching`.

bundlewrap/utils/cmdline.py
```diff
@@ -159,3 +159,32 @@ def get_target_nodes(repo, target_strings, node_workers=None):
         exit(1)
 
     return nodes_matching
+
+
+
+def verify_autoskip_selector(nodes, autoskip_selector):
```
MaZderMind: `autoskip_selector` is a list here, also the fact that this function works with `autoskip_selectors` is already mentioned in the function name. I'd suggest the shorter `selectors` as argument name and would also suggest to pluralize the function name to `verify_autoskip_selectors`.

bundlewrap/utils/cmdline.py
```diff
@@ -159,3 +159,32 @@ def get_target_nodes(repo, target_strings, node_workers=None):
...
+    if not autoskip_selector:
+        return set()
+
+    selectors = {
```
MaZderMind: This looks more like a `selector_used` to me :)

PR #886, opened 2026-01-12
rename cdict/sdict to expected_state/actual_state

### LINKED ISSUE #705: rename cdict and sdict
see #704

CroneKorkN:
rename cdict/sdict to expected_state/actual_state

fixes #705

### DIFF
--- a/bundlewrap/items/__init__.py
+++ b/bundlewrap/items/__init__.py
@@ -828,10 +828,10 @@ class Item:
 
     def display_on_create(self, expected_state):
         """
-        Given a expected_state as implemented above, modify it to better suit
-        interactive presentation when an item is created. If there are
-        any when_creating attributes, they will be added to the expected_state
-        before it is passed to this method.
+        Given an expected_state dict as implemented above, modify it to
+        better suit interactive presentation when an item is created. If
+        there are any when_creating attributes, they will be added to
+        the expected_state before it is passed to this method.
 
         MAY be overridden by subclasses.
         """
@@ -849,8 +849,8 @@ class Item:
 
     def display_on_delete(self, actual_state):
         """
-        Given an actual_state as implemented above, modify it to better suit
-        interactive presentation when an item is deleted.
+        Given an actual_state dict as implemented above, modify it to
+        better suit interactive presentation when an item is deleted.
 
         MAY be overridden by subclasses.
         """
--- a/docs/content/guide/dev_item.md
+++ b/docs/content/guide/dev_item.md
@@ -47,9 +47,9 @@ Create a new file called `/your/bundlewrap/repo/items/foo.py`. You can use this 
         @property
         def expected_state(self):
             """
-            Return a state-dict that describes the expeced state of this item
-            as configured in the repo. Returning `None` instead means that the
-            item should not exist on the node.
+            Return a dict describing the expected state of this item on
+            the node. Returning `None` instead means that the item
+            should not exist.
 
             Implementing this method is optional. The default implementation
             uses the attributes as defined in the bundle.
@@ -59,21 +59,21 @@ Create a new file called `/your/bundlewrap/repo/items/foo.py`. You can use this 
         @property
         def actual_state(self):
             """
-            Return a state-dict that describes the actual state of this item
-            on the node. Returning `None` instead means that the item does not
-            exist on the node.
+            Return a dict describing the actual state of this item on
+            the node. Returning `None` instead means that the item does
+            not exist on the node.
 
-            For the item to validate as correct, the values for all keys in
-            self.expected_state have to match this actual_state.
+            For the item to validate as correct, the values for all keys
+            in self.expected_state have to match this actual_state.
             """
             raise NotImplementedError
 
         def display_on_create(self, expected_state):
             """
-            Given a expected_state as implemented above, modify it to better suit
-            interactive presentation when an item is created. If there are
-            any when_creating attributes, they will be added to the expected_state
-            before it is passed to this method.
+            Given an expected_state dict as implemented above, modify it to
+            better suit interactive presentation when an item is created. If
+            there are any when_creating attributes, they will be added to
+            the expected_state before it is passed to this method.
 
             Implementing this method is optional.
             """
@@ -91,8 +91,8 @@ Create a new file called `/your/bundlewrap/repo/items/foo.py`. You can use this 
 
         def display_on_delete(self, actual_state):
             """
-            Given an actual_state as implemented above, modify it to better suit
-            interactive presentation when an item is deleted.
+            Given an actual_state dict as implemented above, modify it to
+            better suit interactive presentation when an item is deleted.
 
             Implementing this method is optional.
             """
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -448,7 +448,7 @@
 
 2020-10-30
 
-* fixed tomlkit types not being accepted as statedict values
+* fixed tomlkit types not being accepted as actual_state values
```
vain: I wouldn’t touch the CHANGELOG. This is a historical document and should describe things as they were back then.

(Besides, due to the ambiguous nature of those old terms, this change might be incorrect: “statedict” might have referred to the general concept of “this is a dict that describes state”, i.e. *both* `actual_state` and `expected_state`.)

bundlewrap/cmdline/items.py
```diff
@@ -8,7 +8,7 @@
 from ..exceptions import FaultUnavailable
 from ..items import BUILTIN_ITEM_ATTRIBUTES
 from ..utils.cmdline import get_item, get_node
-from ..utils.dicts import statedict_to_json
+from ..utils.dicts import actual_state_to_json
```
vain: Same here, this isn’t correct: “statedict” refers to the general concept of “this is a dict that holds state (either configured or currently on the node)”.

For example, `statedict_to_json()` is called in `bundlewrap/cmdline/items.py` around line 142 and if you scroll up a few lines, you’ll see that its argument can either be a `cdict` or an `sdict`.

I’d suggest keeping the term “statedict” as a general term and `actual_state` and `expected_state` are both statedicts.

docs/content/guide/dev_item.md
```diff
@@ -142,9 +141,9 @@ Create a new file called `/your/bundlewrap/repo/items/foo.py`. You can use this
 Step 3: Implement methods
 -------------------------
 
-You should probably start with `sdict()`. Use `self.run("command")` to run shell commands on the current node and check the `stdout` property of the returned object.
+You should probably start with `actual_state()`. Use `self.run("command")` to run shell commands on the current node and check the `stdout` property of the returned object.
 
-The only other method you have to implement is `fix`. It doesn't have to return anything and just uses `self.run()` to fix the item. To do this efficiently, it may use the provided parameters indicating which keys differ between the should-be sdict and the actual one. Both sdicts are also provided in case you need to know their values.
+The only other method you have to implement is `fix`. It doesn't have to return anything and just uses `self.run()` to fix the item. To do this efficiently, it may use the provided parameters indicating which keys differ between the should-be actual_state and the actual one. Both actual_states are also provided in case you need to know their values.
```
vain: When using underscores in Markdown files, please be careful with underlines, because they signify emphasize. Either write \`actual_state\` in backticks or use a backslashed actual\\\_state.

bundlewrap/cmdline/parser.py
```diff
@@ -501,7 +501,7 @@ def build_parser_bw():
     information_group.add_argument(
         "--state",
```
fkusei: We might want to rename this, so it's clear this is the *actual state*, not the *expected state*.

bundlewrap/items/__init__.py
```diff
@@ -683,17 +682,19 @@ def run(self, command, **kwargs):
...
-        dict). Returning `None` instead means that the item should
-        not exist.
+        Return a dict describing the expected state of this item on the node.
+        Returning `None` instead means that the item should not exist.
 
         MAY be overridden by subclasses.
         """
         return self.attributes
 
+    def cdict(self):
```
fkusei: Why keep this? Do we need this anywhere?

bundlewrap/items/__init__.py
```diff
@@ -878,20 +879,22 @@ def preview(self):
...
 
         For the item to validate as correct, the values for all keys in
-        self.cdict() have to match this statedict.
+        self.expected_state() have to match this actual_state.
 
         MUST be overridden by subclasses.
         """
         raise NotImplementedError()
 
+    def sdict(self):
```
fkusei: See line 694

docs/content/guide/dev_item.md
```diff
@@ -47,9 +44,9 @@ Create a new file called `/your/bundlewrap/repo/items/foo.py`. You can use this
         def __repr__(self):
             return "<Foo attribute:{}>".format(self.attributes['attribute'])
 
-        def cdict(self):
+        def expected_state(self):
             """
-            Return a statedict that describes the target state of this item
+            Return a actual_state that describes the target state of this item
```
fkusei: expected_state

docs/content/guide/dev_item.md
```diff
@@ -158,9 +155,9 @@ Create a new file called `/your/bundlewrap/repo/items/foo.py`. You can use this
 Step 3: Implement methods
 -------------------------
 
-You should probably start with `sdict()`. Use `self.run("command")` to run shell commands on the current node and check the `stdout` property of the returned object.
+You should probably start with `actual_state()`. Use `self.run("command")` to run shell commands on the current node and check the `stdout` property of the returned object.
 
-The only other method you have to implement is `fix`. It doesn't have to return anything and just uses `self.run()` to fix the item. To do this efficiently, it may use the provided parameters indicating which keys differ between the should-be sdict and the actual one. Both sdicts are also provided in case you need to know their values.
+The only other method you have to implement is `fix`. It doesn't have to return anything and just uses `self.run()` to fix the item. To do this efficiently, it may use the provided parameter `status.keys_to_fix` indicating which keys differ between the `expected_state` and the `actual_state`. Both state-dicts are also provided in case you need to know their values.
```
fkusei: The parameter is `status`, which is an `ItemStatus` with attributes `keys_to_fix`, `must_be_created` and `must_be_deleted`. Docs should reflect that.

fkusei (commented):
Thoughts:

- We have a lot of `def expected_state(): expected_state = self.attributes.copy()`. This means we're overwriting the variable inside the function, which works fine, but is difficult to read. Can we maybe rename the inner variable to `state` or something like that? (probably same for `def actual_state()`
- Can we make `expected_state` and `actual_state` a property? This feels "more correct"

bundlewrap/items/__init__.py
MaZderMind: we discussed that with this in place it would not be a very breaking change, because existing custom-items would continue to work. However renaming the methods in them would also not be too big of a burden on users.

docs/content/guide/dev_item.md
MaZderMind: yea, the docs were not to clear to begin with…

bundlewrap/items/__init__.py
Kunsi: huh? This method will never be called by bundlewrap, neither here nor in custom items. It's just dead code.

MaZderMind: either way, the way it was implemented did not achieve this goal either, i'l simply remove them

Kunsi (approved):

vain (approved):

vain:
Ready to merge, imo.

PR #889, opened 2026-01-12
cleanup `bw items` output

### LINKED ISSUE #602: `bw items` output format consistency
`bw items node item` shows JSON, the same command with `--attrs` a table

maybe `--json`?

### LINKED ISSUE #649: `bw items --repr` has inconsistent behaviour
```/home/kunsi/git/bundlewrap|main✓➤ bw items --repr rx300 | grep ZFSPool
<ZFSPool name:tank autoexpand:None autoreplace:None autotrim:True config:[{'type': 'raidz', 'devices': {'/dev/sdc', '/dev/sdd', '/dev/sdb', '/dev/sda'}}]>
/home/kunsi/git/bundlewrap|main✓➤ bw items --repr rx300 zfs_pool:tank
{
    "autotrim": true
}
```
I'd expect `bw items --repr rx300 zfs_pool:tank` to also show `__repr__()` instead of `cdict()`.

MaZderMind:
This PR changes the behavior of the various modes of `bw items` so that their output-format follows a clear pattern and can be selected.

The following output-formats are supported:
```
bw items --format json # json formatted
bw items [--format table] # bw-table, default
BW_TABLE_STYLE=grep bw items [--format table] # greppable
```

The following `bw items` modes are available (this list has not changed and is copied here for better testing, however the output-format has changed for some of them, see below):
```
bw items NODE # list of items, was always a flat list
bw items NODE --repr # list of items was always a flat list
bw items NODE --blame # list of items grouped by bundle, was always table
bw items NODE ITEM # cdict of item, was always json
bw items NODE ITEM --repr # repr of item, did not work before, see #649
bw items NODE ITEM --state # sdict of item, was always json
bw items NODE ITEM --attrs # internal attributes of item, was always table

bw items NODE ITEM --preview # content of file, was and is only printed as-is

bw items NODE --write-file-previews /tmp/some-temp-dir # no formattable output
```

fixes #602
fixes #649

### DIFF
--- a/bundlewrap/cmdline/items.py
+++ b/bundlewrap/cmdline/items.py
@@ -151,11 +151,11 @@ def show_single_item(node, item, representation, args):
 
     elif representation == ItemRepresentation.REPR:
         data = [repr(item)]
-        format_data(data, args['format'], table_headers=[_('item')])
+        format_data(data, args, table_headers=[_('item')])
         return
 
     # print attributes key-value
-    format_data(data, args['format'], table_headers=[_('attribute'), _('value')])
+    format_data(data, args, table_headers=[_('attribute'), _('value')])
 
 
 def show_single_item_preview(node, item):
@@ -203,17 +203,17 @@ def list_all_items(node, args):
         for item in sorted(node.items):
             data.append(repr(item) if show_repr else item.id)
 
-    format_data(data, args['format'], table_headers=table_headers)
+    format_data(data, args, table_headers=table_headers)
 
 
-def format_data(data, fmt, table_headers=None):
+def format_data(data, args, table_headers=None):
     """
     Formats a list or a dict (with scalar or list values) according to the requested format
     """
-    if fmt == 'json':
+    if args['format_json']:
         io.stdout(statedict_to_json(data, pretty=True))
 
-    elif fmt == 'table':
+    else:
         format_data_table(data, table_headers)
 
 
--- a/bundlewrap/cmdline/parser.py
+++ b/bundlewrap/cmdline/parser.py
@@ -505,13 +505,10 @@ bundle:my_bundle  # items in this bundle
         help=_("show actual item status on node instead of should-be configuration"),
     )
     parser_items.add_argument(
-        "-F",
-        "--format",
-        action='store',
-        dest='format',
-        choices=['table', 'json'],
-        default='table',
-        help=_("choose the output formatting"),
+        "--json",
+        action='store_true',
+        dest='format_json',
+        help=_("show output as json instead of the usual table"),
     )
 
     # bw lock
--- a/tests/integration/bw_items.py
+++ b/tests/integration/bw_items.py
@@ -248,7 +248,7 @@ def test_show_auto_needs(tmpdir):
             },
         },
     )
-    stdout, stderr, rcode = run("BW_TABLE_STYLE=grep bw items --attr --format table node1 directory:/foo/bar needs", path=str(tmpdir))
+    stdout, stderr, rcode = run("BW_TABLE_STYLE=grep bw items --attr node1 directory:/foo/bar needs", path=str(tmpdir))
     assert stdout.decode() == """attribute\tvalue
 needs\taction:
 needs\tdirectory:/foo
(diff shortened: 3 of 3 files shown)


bundlewrap/cmdline/items.py
```diff
@@ -11,12 +11,44 @@
 from ..utils.dicts import statedict_to_json
 from ..utils.table import ROW_SEPARATOR, render_table
 from ..utils.text import bold, green, mark_for_translation as _, red, yellow
-from ..utils.ui import io, page_lines
+from ..utils.ui import io, page_lines, TTY
+
+
+class ItemRepresentation(enum.Enum):
+    PREVIEW = 'preview'
```
fkusei, line 18: Do we actually need the values here? If not, we could use `enum.auto()` instead.

bundlewrap/cmdline/items.py
```diff
@@ -28,142 +60,199 @@ def write_preview(file_item, base_path):
...
+    table = [
+        [
+            bold(header)
+            for header in table_headers
+        ]
+    ]
+
+    if isinstance(data, (list, set)):
+        table.append(ROW_SEPARATOR)
+        for v in sorted(data):  # iterate a list
```
fkusei: We should not re-sort lists here (and everywhere else in here).

bundlewrap/cmdline/items.py
```diff
@@ -28,142 +60,199 @@ def write_preview(file_item, base_path):
         f.write(content)
 
 
-def bw_items(repo, args):
-    node = get_node(repo, args['node'])
-    if args['preview'] and not args['item']:
-        io.stderr(_("{x} no ITEM given for preview").format(x=red("!!!")))
+def render_file_previews(node, file_preview_path, args):
+    if args['item']:
+        io.stderr(_("{x} use --file-preview to preview single files").format(x=red("!!!")))
```
fkusei: This option was removed from the parser (see below), use `--preview`

bundlewrap/cmdline/parser.py
```diff
@@ -504,6 +503,14 @@ def build_parser_bw():
         dest='show_sdict',
         help=_("show actual item status on node instead of should-be configuration"),
     )
+    parser_items.add_argument(
+        "-F",
+        "--format",
+        action='store',
+        dest='format',
+        choices=['json', 'table', 'flat'],
+        help=_("choose the output formatting, default is table when stdout is a tty and table when it isn't"),
```
fkusei: "... and flat if it isn't"

tests/integration/bw_items.py
```diff
@@ -248,10 +248,11 @@ def test_show_auto_needs(tmpdir):
             },
         },
     )
-    stdout, stderr, rcode = run("BW_TABLE_STYLE=grep bw items --attr node1 directory:/foo/bar needs", path=str(tmpdir))
+    stdout, stderr, rcode = run("BW_TABLE_STYLE=grep bw items --attr node1 --format table directory:/foo/bar needs", path=str(tmpdir))
+    print(stdout.decode())
     assert stdout.decode() == """attribute\tvalue
 needs\taction:
-needs\tdirectory:/foo
```
fkusei, line 254: Old version was better. We should keep it.

MaZderMind: The old version was however a special hack that only was implemented for this special case:
https://github.com/bundlewrap/bundlewrap/blob/561a1de/bundlewrap/cmdline/items.py#L123

I would rather change the output of `BW_TABLE_STYLE` for all tables instead.

bundlewrap/cmdline/parser.py
MaZderMind: indeed.

bundlewrap/cmdline/parser.py
```diff
@@ -480,30 +479,40 @@ def build_parser_bw():
...
-    parser_items.add_argument(
+    information_group.add_argument(
         "--state",
         action='store_true',
         dest='show_sdict',
         help=_("show actual item status on node instead of should-be configuration"),
     )
+    parser_items.add_argument(
+        "-F",
+        "--format",
```
fkusei: Maybe do `--json` and `--table` instead.

tests/integration/bw_items.py
MaZderMind: Resolved this in a much more generic way in the table-subsystem which can now also accept lists, sets, tuples and mostly any other data-type.

bundlewrap/cmdline/parser.py
MaZderMind: yes, would match the other args much better

bundlewrap/cmdline/items.py
```diff
@@ -28,142 +121,124 @@ def write_preview(file_item, base_path):
...
+def format_data(data, fmt, table_headers=None):
+    """
+    Formats a list or a dict (with scalar or list values) according to the requested format
+    """
+    if fmt == 'json':
+        io.stdout(statedict_to_json(data, pretty=True))
+
+    elif fmt == 'table':
+        format_data_table(data, table_headers)
+
```
Kunsi, line 218: ... else `raise ValueError(fmt)` oder so

MaZderMind: hold on for a second, just pushed that. I wanted to see if the whole table rewrite would test okay with the full test-suite on all python versions first ;)

PR #890, opened 2026-01-12
Item.if

CroneKorkN:


bundlewrap/items/__init__.py
```diff
@@ -336,6 +339,21 @@ def cached_unless_result(self):
         else:
             return False
 
+    @cached_property
+    def cached_if_result(self):
+        """
+        Returns True if 'if' allows this item to run.
+        """
+        if getattr(self, 'if') and (self.ITEM_TYPE_NAME == 'action' or not self.cached_status.correct):
```
fkusei, line 347: Why use `getattr()` here (and below) instead of just `self.if`?

PR #893, opened 2026-01-13
Remove dependency on passlib

### LINKED ISSUE #853: `passlib` is dead, stop using it
https://foss.heptapod.net/python-libs/passlib/-/issues/187

I’ve only seen `bcrypt` break so far, but it’s only a matter of time until the rest fails.

vain:
-   Use `hashlib` directly where appropriate.
-   Use bcrypt/blowfish as a `crypt` algorithm everywhere.
    -   This also removes the `hash_method` attribute from `user` items.

Please see the individual commit messages on how/where this was tested.

Fixes #853.

### DIFF
--- a/docs/content/guide/api.md
+++ b/docs/content/guide/api.md
@@ -323,6 +323,6 @@ Faults also support some rudimentary string operations such as appending a strin
 	>>> repo.vault.password_for("1").b64encode().value
 	'Vk9kNVA='
 	>>> repo.vault.password_for("1").as_htpasswd_entry("username").value
-	'username:$apr1$8be694c7…'
+	'username:$2b$12$MaZ4/O/Kaoy2Corpbb0…'
 
 These string methods are supported on Faults: `format`, `lower`, `lstrip`, `replace`, `rstrip`, `strip`, `upper`, `zfill`


bundlewrap/utils/crypto.py
```diff
@@ -1,19 +1,70 @@
...
+
+    if salt_from is not None and salt is None:
+        # Derive raw bytes from the string that we got. SHA512 is
+        # probably a good choice, as it normalizes the `salt_from`
+        # string to a fixed length.
+        #
+        # The raw salt is 16 bytes long. After "base64" encoding, it is
+        # 22 bytes long.
+        salt = _base64_bcrypt(
+            sha512(salt_from.encode('UTF-8')).digest()[:16]
```
MaZderMind: I don't think it is safe to use only parts of a sha-hash and expect reasonable distributed entropy from it. I think it would be safer to use a dedicated prng here: `random.Random(salt_from.encode('UTF-8')).randbytes(16)` or something thereabouts.

vain: Probably a good idea. Should be changed in other places, too.

vain: This isn’t trivial. I pushed a commit that tries to (somewhat) address this.

bundlewrap/utils/__init__.py
```diff
@@ -10,9 +10,10 @@
 from sys import stderr, stdout
 from tempfile import mkstemp
 
-from passlib.hash import apr_md5_crypt
 from requests import get
 
+from bundlewrap.utils.crypto import b64encode_bcrypt, crypt_bcrypt
```
vain: Use relative import.

bundlewrap/utils/crypto.py
```diff
@@ -1,19 +1,48 @@
...
+
+    # We must now translate from the first alphabet to the second one.
+    t = str.maketrans(alphabet_standard, alphabet_bcrypt)
+    payload_64_bcrypt = payload_b64.translate(t)
+
+    # The padding must also be stripped off.
+    return payload_64_bcrypt.rstrip('=')
+
+
+def crypt_bcrypt(payload, encoding='UTF-8', rounds=12, salt=None):
```
fkusei: s/rounds/cost_factor/ ?

fkusei: Maybe also put it into a constant with `_DEFAULT_BCRYPT_SALT`

bundlewrap/items/postgres_roles.py
```diff
@@ -93,10 +93,12 @@ def sdict(self):
 
     def patch_attributes(self, attributes):
         if 'password' in attributes:
-            attributes['password_hash'] = postgres_context.encrypt(
-                force_text(attributes['password']),
-                user=self.name,
-            )
+            attributes['password_hash'] = 'md5' + md5(
+                (
+                    force_text(attributes['password']) +
```
MaZderMind, line 98: Achtung: `force_text` forced leider nicht wirklich text – ich würde hier noch ein `str` drumrum packen…

bundlewrap/items/users.py
```diff
@@ -285,29 +266,23 @@ def sdict(self):
...
-                )
-            else:
-                attributes['password_hash'] = hash_method.encrypt(
-                    force_text(attributes['password']),
-                    rounds=5000,  # default from glibc
-                    salt=_DEFAULT_SALT if salt is None else salt,
-                )
+
+            attributes['password_hash'] = bcrypt(
+                force_text(attributes['password']),
```
MaZderMind, line 282: Auch hier würde ich ein `str` empfehlen

bundlewrap/items/postgres_roles.py
vain: I have a feeling that we should rather fix `force_text()`. Its docstring says: “Try to return a text aka unicode object from the given data.” Well, it *tries* to do that, but it doesn’t exactly try *hard*. 🥴  (The only case where we truly get a `str` is when you pass `bytes`. Anything else is undefined.)

Given that you have approved the PR and that usage of `force_text()` is not new, I’m going to merge this now and open a new issue for the `force_text()` inconsistencies.

vain:
@fkusei You had marked this as “changes requested”, which blocks the merge. Has anything been addressed? Can we merge?

bundlewrap/items/postgres_roles.py
MaZderMind: Yeah it was more of a soft hint – you know better if this might break hard with non-str objects here.

PR #894, opened 2026-01-13
introduce MetadataUnavailable exception

### LINKED ISSUE #661: raise custom MetadataUnavailable exception from metadata.get()
(in reactors, instead of KeyError)

manueldomke:
fixes #661


bundlewrap/utils/metastack.py
```diff
@@ -1,10 +1,10 @@
+from bundlewrap.exceptions import MetadataUnavailable
```
fkusei: Use relative import

bundlewrap/exceptions.py
```diff
@@ -14,6 +14,13 @@ class FaultUnavailable(Exception):
     pass
 
 
+class MetadataUnavailable(Exception):
```
fkusei: Create `class BundlewrapError(Exception): pass` and make every exception inherit that.

bundlewrap/utils/metastack.py
```diff
@@ -50,7 +50,7 @@ def get(self, path):
                         result = merge_dict({'data': value}, result)
 
         if undef:
-            raise KeyError('/'.join(path))
+            raise MetadataUnavailable('/'.join(path))
```
fkusei: Hand over `path` here and add `def __str__(self): return '/'...` to exception

fkusei: https://github.com/bundlewrap/bundlewrap/blob/0e202974739b4c5c6dcc137985c9ac8ceea5f882/bundlewrap/deps.py#L11

fkusei (changes requested):

PR #895, opened 2026-01-13
add exception for taget strings that don't resolve

### LINKED ISSUE #869: UX degradation in 4.24.0
In #846 we reworked the code to have a public API for the then-existing `bundlewrap.utils.cmdline.get_target_nodes()`.

If you called 4.23.x like this: `bw apply node1 node2 node_doesnt_exist`
bundewrap did tell you `No such Node or Group: node_doesnt_exist`

With the new implementation, those errors get eaten silently. In the above example, `node_doesnt_exist` is simply ignored, which could lead to users not applying all the systems they intended to. If they only type `bw apply node_doesnt_exist`, only then they get a rather not helpful message of `Input did not match any nodes`.

sschiller-seibert:
An attempt to fix the ux degradation mentioned in #869 by adding an explicit exception that gets raised when no target can be found for a given string.

fixes #869

### DIFF
--- a/docs/content/guide/api.md
+++ b/docs/content/guide/api.md
@@ -149,6 +149,8 @@ The following expressions can be used:
 ` `!group:my_group` all nodes not in this group
 - `"lambda:node.metadata_get('foo/magic', 47) < 3"` all nodes whose `metadata["foo"]["magic"]` is less than three
 
+Throws the exception `NoSuchTarget` if one of the expressions in the target string doesn't resolve to nodes.
+
 <br>
 
 ### bundlewrap.node.Node()


bundlewrap/repo.py
```diff
@@ -701,7 +702,7 @@ def nodes_matching(self, target_strings, lambda_workers=None):
                         group = self.get_group(name)
                         targets.update(group.nodes)
                     except NoSuchGroup:
-                        continue
+                        raise NoSuchTarget(_(f"Target string {name} does match neither bundle, nor group, node or lambda."))
```
fkusei: only hand over `name`, deal with formatting in exception or somewhere else.

sschiller-seibert: [d6df518](https://github.com/bundlewrap/bundlewrap/pull/895/commits/d6df518efb2eeb607b0bbfcfe797edb9879eba0f) should resolve this

PR #897, opened 2026-01-13
Removed the not used can_login in postgres_roles and its test.

simonbrielbeck-seibert:
Removed attribute `can_login` in items/postgres_roles.py


bundlewrap/items/postgres_roles.py
```diff
@@ -44,7 +43,7 @@ def get_role(node, role):
         else:
             role_attrs[AUTHID_COLUMNS[key]] = value
 
-    for bool_attr in ('can_login', 'superuser'):
+    for bool_attr in ('superuser'):
```
fkusei:
```suggestion
    for bool_attr in ('superuser',):
```
Otherwise we're doing that for each character in the string.

PR #900, opened 2026-01-13
remove `Item.get_auto_deps()`

### LINKED ISSUE #899: remove `Item.get_auto_deps()`
do it with #898, we need to touch all items anyway

fkusei:
this fixes #899

### DIFF
--- a/bundlewrap/items/files.py
+++ b/bundlewrap/items/files.py
@@ -435,11 +435,15 @@ class File(Item):
                     ))
                 else:
                     deps.add(item.id)
+            elif item.ITEM_TYPE_NAME == "zfs_dataset":
+                if item.attributes['mountpoint'] not in (None, "none"):
+                    if is_subdirectory(item.attributes['mountpoint'], self.name):
+                        deps.add(item.id)
             elif item.ITEM_TYPE_NAME in ('directory', 'symlink'):
                 if is_subdirectory(item.name, self.name):
                     deps.add(item.id)
         return {
-            'needs': deps,
+            'after': deps,
         }
 
     def sdict(self):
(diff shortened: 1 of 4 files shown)


bundlewrap/items/files.py
```diff
@@ -434,11 +434,13 @@ def get_auto_deps(self, items):
...
                 else:
-                    deps.append(item.id)
+                    deps.add(item.id)
             elif item.ITEM_TYPE_NAME in ('directory', 'symlink'):
                 if is_subdirectory(item.name, self.name):
-                    deps.append(item.id)
-        return deps
+                    deps.add(item.id)
+        return {
+            'after': deps,
```
vain: Why is this `after` and all the other ones are `needs`? 🤔

bundlewrap/items/pkg_pamac.py
```diff
@@ -46,12 +46,11 @@ def pkg_installed(self):
...
         self.run("pamac remove --no-confirm --unneeded --orphans {}".format(quote(self.name)), may_fail=True)
 
-    def get_auto_deps(self, items):
-        deps = []
+    def get_auto_attrs(self, items):
         for item in items:
             if item == self:
                 continue
-            if item.ITEM_TYPE_NAME in ("pkg_pacman") and item.name == self.name:
+            if item.ITEM_TYPE_NAME in ("pkg_pacman",) and item.name == self.name:
```
MaZderMind, line 53: It looks to me if this is not the right method for this kind of check – it does not actually generate attrs, just mis-uses this hook point to do additional check. I'd like to open a ticket to clean this up (does not have to be fixed in this pr) -> #911

PR #901, opened 2026-01-13
HooksProxy: show users where errors occur

fkusei:


vain:
@fkusei Can you give an example of what this changes? 🤔

For example, when I define a hook that raises an exception, then I get the same output as before:

```python
def node_apply_start(repo, node, interactive=False, **kwargs):
    raise Exception('foo')
```
```
$ bw apply -i mynode
  File "/home/foobar/git/bundlewrap/bundlewrap/concurrency.py", line 137, in run
    result = self._get_result()
  File "/home/foobar/git/bundlewrap/bundlewrap/concurrency.py", line 77, in _get_result
    raise exception
  File "/usr/lib/python3.14/concurrent/futures/thread.py", line 86, in run
    result = ctx.run(self.task)
  File "/usr/lib/python3.14/concurrent/futures/thread.py", line 73, in run
    return fn(*args, **kwargs)
  File "/home/foobar/git/bundlewrap/bundlewrap/node.py", line 765, in apply
    self.repo.hooks.node_apply_start(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        repo=self.repo,
        ^^^^^^^^^^^^^^^
        node=self,
        ^^^^^^^^^^
        interactive=interactive,
        ^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/foobar/git/bundlewrap/bundlewrap/repo.py", line 211, in hook
    self.__module_cache[filename][event](**kwargs)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
  File "/home/foobar/git/isac/bw/hooks/test_ssl.py", line 11, in node_apply_start
    raise Exception('foo')

Exception('foo')
! mynode  foo

!!! There was an error, repeated below.

! mynode  foo
```

fkusei:
This has helped yesterday by telling users which hook used wrong variable names. (Probably also other errors as well)

vain (approved):

PR #903, opened 2026-01-14
[WIP] Changelog and migration guide for 5.0

vain:


CHANGELOG.md
```diff
@@ -1,3 +1,18 @@
...
+TODO date
+
+* a user's identity now automatically includes the Git branch name (BACKWARDS INCOMPATIBLE)
+* support for `svc_upstart` items has been removed (BACKWARDS INCOMPATIBLE)
+* hashing (files, node hashes, ...) now uses sha256 instead of sha1 (BACKWARDS INCOMPATIBLE)
+* `BW_SCP_ARGS` no longer defaults to the value of `BW_SSH_ARGS`, because these two programs use different sets of arguments (BACKWARDS INCOMPATIBLE)
+* exit with an error if `-s` selectors don't match anything (BACKWARDS INCOMPATIBLE)
+* `repo.nodes_matching()` throws `NoSuchTarget` when expressions don't match anything (BACKWARDS INCOMPATIBLE)
+* the "Running hooks ..." line is more verbose, bw calls your hooks differently (BACKWARDS INCOMPATIBLE)
+* the `bundles/` directory is required to be present for bw to consider a directory a bw repository
```
vain, line 26: Should we also classify this as a breaking change? It’s more of a bugfix, I think, because `repo.populate_from_path()` always assumed that `bundles/` is present, i.e. you couldn’t actually use bw without that directory.

The only change is that `repo.is_repo()` now properly fails.

fkusei: `Repository.is_repo()` is not listed in https://docs.bundlewrap.org/guide/api/ so i'd consider it a private API, so we don't need to list it as breaking change in my opinion.

docs/content/guide/migrate_45.md
```diff
@@ -0,0 +1,75 @@
...
+
+## `repo.nodes_matching()` throws `NoSuchTarget` when expressions don't match anything
+
+You will have to catch this exception and decide what to do. Previously, this case was silently ignored.
+
+<br>
+
+## Hooks are called with positional arguments
+
+Hooks are called with position arguments now. The names in your function signature must match the ones in [our documentation](../repo/hooks.md).
```
fkusei: With *named* arguments. We got rid of positional arguments.

vain: Errr, right. Thinko.

CHANGELOG.md
```diff
@@ -1,3 +1,20 @@
...
+* `BW_SCP_ARGS` no longer defaults to the value of `BW_SSH_ARGS`, because these two programs use different sets of arguments (BACKWARDS INCOMPATIBLE)
+* exit with an error if `-s` selectors don't match anything (BACKWARDS INCOMPATIBLE)
+* `repo.nodes_matching()` throws `NoSuchTarget` when expressions don't match anything (BACKWARDS INCOMPATIBLE)
+* the "Running hooks ..." line is more verbose, bw calls your hooks differently (BACKWARDS INCOMPATIBLE)
+* items: `get_auto_deps()` has been removed, always use `get_auto_attrs()` now (BACKWARDS INCOMPATIBLE)
+* dependency on passlib has been removed, this affects several hashing algorithms (BACKWARDS INCOMPATIBLE)
+* the `bundles/` directory is required to be present for bw to consider a directory a bw repository
+* fixed mkdocs warnings
+* removed unused attribute `can_login` from `postgres_roles.py`
+
```
MaZderMind, line 31: * the output of `bw items` war harmonized over all subcomannds. Some subcommands that previously generated json now default to table-output and will need `--json` to switch back to json output.
* Tables printed to the cli with only one column are now formatted as flat list without decorators.
* Tables printed to the cli with The Format of `BW_TABLE_STYLE=grep` were changed to repeat literal-columns for every row of an array for all tables (not just some), this might require changes to your scripts if you parse `bw` command-output.
* `bw item NODE ITEM --state` is now called `--actual-state` to clarify its function

manueldomke, line 31: - Introduce BundlewrapError as the new base exception class for the project to standardize error handling.
- Add MetadataUnavailable exception to explicitly handle scenarios where metadata cannot be retrieved yet (fixes #661).

CHANGELOG.md
```diff
@@ -1,3 +1,20 @@
...
 # 4.24.0
```
MaZderMind, line 32: `node.metadata` does not behave like a dict anymore, especially `node.metadata.items()`, `node.metadata.keys()`, `node.metadata.values()` and dict-style access `node.metadata[…]` is not possible anymore. Also `node.metadata.get('key')` does now throw a `MetadataUnavailable` instead of silently returning `None` in the same way that `node.metadata.get('key/other')` always did. `.get(…)` is now the only supported api to access metadata.

CHANGELOG.md
```diff
@@ -1,3 +1,20 @@
...
```
MaZderMind, line 33: internally `cdict` and `sdict` was renamed to `expected_state` and `actual_state`. This is mostly relevant for your custom items which need to be changed accordingly. Also the `expected_state` and `actual_state` methods of classes extending from `Item` now must beannotated with `@property`.

MaZderMind (approved):

CHANGELOG.md
vain: (Note to self: Not yet merged.)

CHANGELOG.md
vain: (Note to self: Not yet merged.)

vain: (Note to self: Include this now.)

docs/content/guide/migrate_45.md
```diff
@@ -0,0 +1,265 @@
...
+### `MetadataUnavailable` instead of `KeyError`
+
+[Metadata reactors](../repo/metadata.py.md#reactors) must now throw `MetadataUnavailable` to indicate that metadata is not available. For example in `metadata.py`:
+
+    from bundlewrap.exceptions import MetadataUnavailable
+
+    @metadata_reactor
+    def foo(metadata):
+        if some_condition:
+            raise MetadataUnavailable()
```
MaZderMind: I would love a more realistic example here and an explanation *when* reactors need to throw this exception and that before a KeyError was enough. We should also mention that simply accessing metadata.get will automatically do the right thing.

docs/content/guide/migrate_45.md
```diff
@@ -0,0 +1,265 @@
...
+
+This has been deprecated for a long time. If you still have dict-style access like this:
+
+    version = metadata['my_program']['version']
+
+Then you must replace it with this, because this is the only public API of these objects now:
+
+    version = metadata.get('my_program/version')
+
+Also, `metadata.items()`, `metadata.keys()`, and `metadata.values()` is gone.
```
MaZderMind: We should mention here that is is possibly to get all metadata as a dict by calling `.get(tuple())`.

docs/content/guide/migrate_45.md
```diff
@@ -0,0 +1,265 @@
...
+<br>
+
+### `bw apply` and `bw verify`: `-s` without matches is an error now
+
+This is an error now:
+
+    $ bw verify hw.switch-foobar -s tag:causes-downtime
+    !!! the following selectors for --skip do not match any items: tag:causes-downtime
+
+The intention is to catch typos. For example, `-s tag:causes_downtime` previously went unnoticed and might have unintentionally restarted some services.
```
MaZderMind: Is there a typo in the tag? I can't see one but would have expected it as an example.

docs/content/guide/migrate_45.md
```diff
@@ -0,0 +1,265 @@
...
+
+### `bw plot --no-depends-reverse` has been removed
+
+Instead, reverse dependencies are shown with dashed lines.
+
+<br>
+
+### `bw test` always fails if `test_with` fails
+
+When a `test_with` command failed, we previously ignored this *if* its exit code was 126, 127, or 255. This is no longer the case.
```
MaZderMind, line 319: We should explain that these error codes relate to the command missing on the dev's machine and that we now require the commands to be present for be test to succeed.

docs/content/guide/migrate_45.md
```diff
@@ -0,0 +1,265 @@
...
+
+### `bw lock -i $selector`: Verify if `$selector` matches
+
+To prevent typos and accidents, BundleWrap will now verify if these item selectors match anything. If you don't want this, use `--skip-item-verification`.
+
+<br>
+
+### `bw lock add`: Warning when there are locks always uses a pager
+
+`bw lock add` shows a message like `Your lock was added, but the node was already locked by ...`, followed by a list of existing locks. This output is always piped through a pager now.
```
MaZderMind: "Always" is not true, the pager is skipped if stdout is not a tty

docs/content/guide/migrate_45.md
fkusei: I think this is related to dash vs. underscore. I'd rather have a more obvious example here as well.

MaZderMind: Ohhh I did totally miss that. Maybe adding ("with an underscore instead of a dash") would be enough to clarify the example.

docs/content/guide/migrate_45.md
Kunsi: Something like that, i think:

> Previously, you could write stuff like this:
> ```python
> @metadata_reactor
> def my_reactor(metadata):
>     for name, config in metadata.get('something', {}).items():
>         # key 'ips' may or may not (yet) be available here
>         ips = config['ips']
> ```
>
> If 'ips' was not available, the raised `KeyError` would have been caught by BundleWrap and had triggered a reactor re-run at a later point. This is no longer the case.
>
> Instead, we recommend writing code like this:
> ```python
> @metadata_reactor
> def my_reactor(metadata):
>     for name in metadata.get('something', {}):
>         # key 'ips' may or may not (yet) be available here
>         ips = metadata.get(('something', name, 'ips'))  # or .get(f'something/{name}/ips')
> ```
>
> Doing it this way will ensure BundleWrap will correctly know which paths were requested, but not yet available and trigger reactor runs accordingly.

We should not recommend users to raise `MetadataUnavailable` in reactors themselves.

vain: Yes, that’s better, thank you.

docs/content/guide/migrate_45.md
vain: Okay, fixed.

Do your changes *only* affect `bw lock`? Are there other places that might block now?

MaZderMind: I did not change that, that is how `page_line` always(tm) worked: https://github.com/bundlewrap/bundlewrap/blob/main/bundlewrap/utils/ui.py#L115

My comment was only regarding this line in the changelog.

CHANGELOG.md
```diff
@@ -1,3 +1,34 @@
...
+* canned actions inherit tags (BACKWARDS INCOMPATIBLE)
+* `bw test` no longer ignores some exit status codes of `test_with` (BACKWARDS INCOMPATIBLE)
+* item selectors in `bw lock -i ...` are now verified and raise an error if they match nothing (BACKWARDS INCOMPATIBLE)
+* `MetadataUnavailable` is raised instead of `KeyError` when `metadata.get()` cannot find the requested metadata key (BACKWARDS INCOMPATIBLE)
+* `Item.display_dicts()` replaced by `Item.display_on_fix()` and the `keys` argument is a set now (BACKWARDS INCOMPATIBLE)
+* `bw lock add`'s warning when there already are locks always uses a pager now (BACKWARDS INCOMPATIBLE)
+* `metadata` objects no longer try to behave like dicts, you must use `get()` now (BACKWARDS INCOMPATIBLE)
+* terminology: `cdict` and `sdict` have been renamed to `expected_state` and `actual_state`, and they need `@property` now; this affects custom items (BACKWARDS INCOMPATIBLE)
+* metadata reactors now raise an exception when they do not read any metadata; users should use `defaults` instead (BACKWARDS INCOMPATIBLE)
+* `directory` items auto-depend on corresponding `zfs_dataset` items; users should remove manual dependencies
```
fkusei: ... and `files`

vain: And files.

PR #908, opened 2026-01-14
extend docs

MaZderMind:
Document subcommands of `bw` and fix rendering.errors in api-docs.

### DIFF
--- a/docs/content/guide/cli.md
+++ b/docs/content/guide/cli.md
@@ -94,6 +94,24 @@ It offers an interesting view into the internal complexities BundleWrap has to d
 
 ## bw stats
 
+Show some statistics about your repository.
+
+```
+$ bw stats
+╭───────┬───────────────────╮
+│ count │ type              │
+├───────┼───────────────────┤
+│     1 │ nodes             │
+│     0 │ groups            │
+│     1 │ bundles           │
+│     0 │ metadata defaults │
+│     0 │ metadata reactors │
+│     2 │ items             │
+├───────┼───────────────────┤
+│     2 │ file              │
+╰───────┴───────────────────╯
+```
+
 ## bw test
 
 ```none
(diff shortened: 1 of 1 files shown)


docs/content/guide/cli.md
```diff
@@ -6,51 +6,98 @@ The `bw` utility is BundleWrap's command line interface.
...
+```none
+$ bw plot node mynode | dot -Tsvg -omynode.svg</code></pre>
+```
 
 You won't be using this every day, but it's pretty cool. The above command will create an SVG file (you can open these in your browser) that shows the item dependency graph for the given node. You will see bundles as dashed rectangles, static dependencies (defined in BundleWrap itself) in green, auto-generated dependencies (calculated dynamically each time you run `bw apply`) in blue and dependencies you defined yourself in red.
 
 It offers an interesting view into the internal complexities BundleWrap has to deal with when figuring out the order in which your items can be applied to your node.
 
-<br>
+## bw stats
```
Kunsi, line 95: Needs docs

docs/content/guide/cli.md
```diff
@@ -65,6 +112,133 @@ It offers an interesting view into the internal complexities BundleWrap has to d
...
+d063cfc437616f7753accf990e43ceece4b2fb92e51ea0314de0effc6ae1fa81
+
+$ bw hash ubuntu-24.04
+51e5099152418033e002bace2a34d612815df759b18b15106cb462af1b42bf17
+
+$ bw hash mynode user:root
+09b3124df67ecaebae1d740f9985c3bcfa62492cf114b23026028cf1eb457c75
+```
+
+## bw generate-completions
```
Kunsi, line 260: Needs hint that this command is only available if `argcomplete` was manually installed: https://github.com/bundlewrap/bundlewrap/blob/main/bundlewrap/cmdline/parser.py#L33

Kunsi (changes requested):

docs/content/guide/cli.md
MaZderMind: Ah yes I started the command to see what it would produce and it did not finish before i forgot what I was doing 🫣

MaZderMind: _1 hour later it still runs on our prod-repo but I copied the output from the unit-test to the docs. good enough._

PR #910, opened 2026-01-14
docs/content/items: add general documentation for package management items

MaZderMind:
…and add our suggestion for a package management bundle.

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -43,6 +43,7 @@ nav:
   - git_deploy: items/git_deploy.md
   - group: items/group.md
   - k8s_*: items/k8s.md
+  - pkg_*: items/pkg.md
   - pkg_apk: items/pkg_apk.md
   - pkg_apt: items/pkg_apt.md
   - pkg_dnf: items/pkg_dnf.md


docs/content/items/pkg_apk.md
```diff
@@ -1,5 +1,7 @@
 # apk package items
 
+<div class="alert alert-info">See the general remarks on <a href="/items/pkg">package management items</a>.</div>
```
Kunsi, line 3: Can we have these links resolved by mkdocs instead? This would make it easier to later change paths, if needed.

docs/content/items/pkg.md
Kunsi, line 1: Needs to get added to the mkdocs config as well so it gets built

docs/content/items/pkg_apk.md
MaZderMind: I don't think we can while keeping the alert-box. Markdown links don't work in the html block.

docs/content/items/pkg.md
MaZderMind: Ah I wasn't sure about that because it actually did work in `mkdocs serve` but ok, will add it.

Kunsi: I think it will get built but not added to the navigation menu or something like that.

MaZderMind: Yes, that was what I found and I actually did that intentionally, but I can see that this does look like a mistake. I'll see if I can add the document as `pkg_*` to the navbar.

fkusei (approved):

Kunsi (approved):

PR #912, opened 2026-01-14
introduce MetadataUnavailable exception and BundlewrapError as base exception class

### LINKED ISSUE #661: raise custom MetadataUnavailable exception from metadata.get()

manueldomke:
fixes #661


bundlewrap/utils/dicts.py
```diff
@@ -5,6 +5,8 @@
 
 from tomlkit import document as toml_document
 
+from bundlewrap.exceptions import MetadataUnavailable
```
Kunsi: Please use relative imports

bundlewrap/utils/metastack.py
```diff
@@ -1,3 +1,4 @@
+from bundlewrap.exceptions import MetadataUnavailable
```
Kunsi: Also relative import please

bundlewrap/exceptions.py
```diff
@@ -7,56 +13,69 @@ def __init__(self, obj):
...
+    """
+    Raised when a Metadata key cannot be resolved.
+    """
+    def __init__(self, path=None):
+        if path is None:
+            path = {}
+        self.path = path
+        super().__init__(f"Metadata key path unavailable: {'/'.join(self.path)}")
+
+    pass
```
Kunsi: We don't need `pass` here, this is only needed when there's nothing in the class. Since we re-define `__init__()`, we can omit this.

bundlewrap/exceptions.py
```diff
@@ -65,28 +84,28 @@ def __init__(self, target):
         self.target = target
 
     def __repr__(self):
-        return "<NoSuchTarget {}>".format(self.target)
+        return f"<NoSuchTarget {self.target}>"
 
     def __str__(self):
         return self.target
     pass
```
Kunsi: We could remove this `pass` here as well

bundlewrap/metagen.py
```diff
@@ -262,15 +262,15 @@ def _build_node_metadata(self, initial_node_name):
                 break
             elif only_keyerrors:
                 if set(self._reactors_triggered.keys()).difference(reactors_run):
-                    io.debug("all reactors raised KeyErrors, but new ones were triggered")
+                    io.debug("all reactors raised MetadataUnavailables, but new ones were triggered")
```
Kunsi: `MetadataUnavailable` (remove the `s`), also below

bundlewrap/metagen.py
```diff
@@ -372,7 +372,7 @@ def __reactors_to_run(self):
             yield (
                 reactor_id,
                 f"running reactor {reactor_id} because "
-                f"it previously raised a KeyError for: {path_exc[0]}"
+                f"it previously raised a MetadataUnavailable for: {path_exc[0]}"
```
Kunsi: "raised MetadataUnavailable" (omit the `a`)

bundlewrap/metagen.py
```diff
@@ -404,7 +404,7 @@ def __run_reactors(self):
 
         return reactors_run, only_keyerrors
 
-    def __run_reactor(self, node, reactor_name, reactor):
+    def __run_reactor(self, node, reactor_name, reactor): #skipcq: PY-R1000
```
Kunsi: Nitpick:
```python
    def __run_reactor(self, node, reactor_name, reactor):  # skipcq: PY-R1000
```
(two spaces before, one after `#`)

bundlewrap/metagen.py
```diff
@@ -415,32 +415,29 @@ def __run_reactor(self, node, reactor_name, reactor):
         try:
             new_metadata = reactor(node.metadata)
         except KeyError as exc:
```
Kunsi: Is there any reason to catch `KeyError`s specifically? Wouldn't it be better to wrap the call in line 416 inside a `with error_context(node=node.name, reactor=reactor_name):` to give context for all exceptions instead?

bundlewrap/exceptions.py
MaZderMind: Please remove the `pass`, instead implement `__str__` and `__repr__`. I'd suggest to make `path` a required arg, alternatively if this does not work I'd prefer to save `None` and handle this case in `__str__`/`__repr__`.

bundlewrap/exceptions.py
```diff
@@ -150,15 +169,15 @@ class TemplateError(RepositoryError):
...
 
 
-class NodeLockedException(Exception):
+class NodeLockedException(BundlewrapError):
     """
     Raised when a node is already locked during an 'apply' run.
     """
-    pass
+    pass
```
MaZderMind: Final New-Line

bundlewrap/metagen.py
```diff
@@ -91,7 +91,7 @@ def covers(self, candidate_path):
         """
         try:
             return self._covers_cache[candidate_path]
-        except KeyError:
+        except (KeyError, MetadataUnavailable):
```
MaZderMind: Is that right? Isn't a `KeyError` just any generic exception for us now? Why handle it specially here?

bundlewrap/metagen.py
```diff
@@ -415,32 +415,29 @@ def __run_reactor(self, node, reactor_name, reactor):
         try:
             new_metadata = reactor(node.metadata)
         except KeyError as exc:
+            io.stderr(_(
```
MaZderMind: Isn't KeyError just another Exception for us now? Why should we catch and log it here and not just let it pass like we do with any other Exception?

bundlewrap/metagen.py
```diff
@@ -513,4 +512,4 @@ def __run_reactor(self, node, reactor_name, reactor):
                 io.debug(f"rerun of {triggered_reactor} triggered by {self._current_reactor}")
                 self._reactors_triggered[triggered_reactor].add(self._current_reactor)
         else:
-            io.debug(f"{self._current_reactor} returned same result")
+            io.debug(f"{self._current_reactor} returned same result")
```
MaZderMind: Final New-Line

bundlewrap/utils/dicts.py
```diff
@@ -59,7 +61,7 @@ def diff_keys(dict1, dict2):
         try:
             if dict1[key] != dict2[key]:
                 differing_keys.add(key)
-        except KeyError:
+        except (KeyError, MetadataUnavailable):
```
MaZderMind: Ad above, this smells wrong.

bundlewrap/utils/dicts.py
```diff
@@ -383,7 +385,7 @@ def normalize_dict(dict_obj, types):
     for key, value in dict_obj.items():
         try:
             normalize = types[key]
-        except KeyError:
+        except (KeyError, MetadataUnavailable):
```
MaZderMind: Ad above, this smells wrong.

bundlewrap/utils/dicts.py
```diff
@@ -510,6 +514,6 @@ def value_at_key_path(dict_obj, path):
         nested_dict = dict_obj[path[0]]
         remaining_path = path[1:]
         if remaining_path and not isinstance(nested_dict, dict):
-            raise KeyError("/".join(path))
+            raise MetadataUnavailable(path)
         else:
-            return value_at_key_path(nested_dict, remaining_path)
+            return value_at_key_path(nested_dict, remaining_path)
```
MaZderMind: Final New-Line

bundlewrap/utils/metastack.py
```diff
@@ -36,7 +37,7 @@ def get(self, path):
             for layer in reversed(list(partition.values())):
                 try:
                     value = value_at_key_path(layer, path)
-                except KeyError:
+                except (KeyError, MetadataUnavailable):
```
MaZderMind: Ad above, this smells wrong.

bundlewrap/utils/metastack.py
```diff
@@ -87,7 +88,7 @@ def as_blame(self):
     def pop_layer(self, partition_index, identifier):
         try:
             return self._partitions[partition_index].pop(identifier)
-        except (KeyError, IndexError):
+        except (MetadataUnavailable, KeyError, IndexError):
```
MaZderMind: Probably should remove KeyError here, too

bundlewrap/utils/metastack.py
```diff
@@ -97,4 +98,4 @@ def set_layer(self, partition_index, identifier, new_layer):
     def cache_partition(self, partition_index):
         self._cached_partitions[partition_index] = {
             'merged layers': self.as_dict(partitions=[partition_index]),
-        }
+        }
```
MaZderMind: Final New-Line

tests/unit/metastack.py
```diff
@@ -235,4 +236,4 @@ def test_as_blame():
         ('something', 'a_list'): ['base', 'overlay'],
         ('something', 'a_value'): ['base'],
         ('something', 'another_value'): ['unrelated'],
-    }
+    }
```
MaZderMind: Final New-Line

bundlewrap/metagen.py
manueldomke: Proper solution is to catch KeyError like before, because self._covers_cache ist just a regular dict.

bundlewrap/metagen.py
```diff
@@ -414,33 +414,20 @@ def __run_reactor(self, node, reactor_name, reactor):
...
-                f"{self._reactors_with_keyerrors[self._current_reactor]}"
-            )
+                io.debug(f"{self._current_reactor} raised MetadataUnavailable for {exc.path}")
             return False
         except DoNotRunAgain:
             self._reactors[self._current_reactor]['raised_donotrunagain'] = True
             # clear any previously stored exception
-            with suppress(KeyError):
-                del self._reactors_with_keyerrors[self._current_reactor]
+            with suppress(MetadataUnavailable):
```
fkusei: `KeyError` was correct here, because we're working on a regular `dict` here. (Also, we need either this `with suppress()` or the `if` below, not both)

bundlewrap/metagen.py
```diff
@@ -456,15 +443,17 @@ def __run_reactor(self, node, reactor_name, reactor):
             raise exc
         finally:
             self._in_a_reactor = False
-            with suppress(KeyError):
-                del self._reactors_triggered[self._current_reactor]
+            with suppress(MetadataUnavailable):
```
fkusei: Same as 428

bundlewrap/metagen.py
```diff
@@ -456,15 +443,17 @@ def __run_reactor(self, node, reactor_name, reactor):
...
+                if self._current_reactor in self._reactors_triggered:
+                    del self._reactors_triggered[self._current_reactor]
             for path in self._current_reactor_newly_requested_paths:
                 for needed_reactor in self._trigger_reactors_for_path(path, self._current_reactor):
                     self._reactors[needed_reactor]['trigger_on_change'].add(self._current_reactor)
 
         # reactor terminated normally, clear any previously stored exception
-        with suppress(KeyError):
-            del self._reactors_with_keyerrors[self._current_reactor]
+        with suppress(MetadataUnavailable,KeyError):
```
fkusei: Same as 428

bundlewrap/exceptions.py
```diff
@@ -7,56 +13,73 @@ def __init__(self, obj):
...
+class MetadataUnavailable(BundlewrapError):
+    """
+    Raised when a Metadata key cannot be resolved.
+    """
+    def __init__(self, path=None):
+        self.path = path
+
+    def __str__(self):
+        if self.path is None:
+            return "Metadata unavailable"
```
MaZderMind: Hier wäre ein String wie `path: unknown` glaub ich schlüssiger

bundlewrap/metagen.py
```diff
@@ -514,3 +503,4 @@ def __run_reactor(self, node, reactor_name, reactor):
                 self._reactors_triggered[triggered_reactor].add(self._current_reactor)
         else:
             io.debug(f"{self._current_reactor} returned same result")
+
```
MaZderMind: Newlines are still not right everywhere

bundlewrap/metagen.py
manueldomke: there is already a generic Exception being caught in line 433 - I have however removed the redundant except KeyError block as suggested

Kunsi (approved):

Issue #863, opened 2025-11-04
Format output of `bw metadata` as toml


sschiller-seibert:
Now that toml has become the de facto standard for nodefiles, a lot of mental yoga is needed to transate back and forth between the json-like representation in `bw metadata` and toml based nodefiles. Wouldn't it be easier to directly output toml when looking a metadata? Having this behind a feature flag like `-t, --toml` would also be fine.

PR #915, opened 2026-01-14
add toml output to bw metadata

sschiller-seibert:
first draft for #863

### DIFF
--- a/bundlewrap/cmdline/metadata.py
+++ b/bundlewrap/cmdline/metadata.py
@@ -180,7 +180,8 @@ def bw_metadata(repo, args):
             metadata_sorted = _sort_dict_colorblind(metadata)
             if args["toml"]:
                 page_lines(
-                    metadata_to_toml(
+                    force_text(line).replace("\\u001b", "\033")
+                    for line in metadata_to_toml(
                         metadata_sorted,
                         resolve_faults=args['resolve_faults'],
                         sort_keys=False,
--- a/bundlewrap/cmdline/metadata.py
+++ b/bundlewrap/cmdline/metadata.py
@@ -3,7 +3,7 @@ from decimal import Decimal
 from sys import exit
 
 from ..exceptions import MetadataUnavailable
-from ..metadata import metadata_to_json
+from ..metadata import metadata_to_json, metadata_to_toml
 from ..utils import Fault, list_starts_with
 from ..utils.cmdline import exit_on_keyboardinterrupt, get_target_nodes
 from ..utils.dicts import (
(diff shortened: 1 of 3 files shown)


bundlewrap/cmdline/metadata.py
```diff
@@ -175,14 +175,23 @@ def bw_metadata(repo, args):
 
             # now we need to recreate the dict, sorting the keys as if
             # they were not colored (otherwise we'd end up sorted by
-            # color)
+            # co863lor)
```
Kunsi: oops

bundlewrap/metadata.py
```diff
@@ -330,6 +333,28 @@ def metadata_to_json(metadata, resolve_faults=True, sort_keys=True):
...
+def metadata_dict_to_toml(dict_obj, resolve_faults=False):
+    toml_doc = toml_document()
+    for key, value in dict_obj.items():
+        if isinstance(value, tuple):
+            toml_doc[key] = list(value)
+        elif isinstance(value, set):
+            toml_doc[key] = sorted(value)
+        elif isinstance(value, dict):
+            toml_doc[key] = metadata_dict_to_toml(value, resolve_faults)
+        elif isinstance(value, Fault) and resolve_faults:
```
Kunsi: Wouldn't it be better to do this?
```python
elif isinstance(value, Fault):
    if resolve_faults:
        toml_doc[key] = value.value
    else:
        toml_doc[key] = yellow(value._repr_first())
```
This saves one `isinstance()` call per Fault - probably not a lot, but still.

Kunsi (commented):
In general: LGTM

bundlewrap/metadata.py
sschiller-seibert: Good idea.

sschiller-seibert:
Ok, I managed to improve the output quoting and colorization, but I'm unsure if the new function `dumps_ansi` should live where it currently is.

Screenshot from 2026-01-14 17-39-19

Edit: This doesn't output valid toml anymore. Needs further optimization or a revert to the previous state.

bundlewrap/metadata.py
```diff
@@ -330,6 +334,30 @@ def metadata_to_json(metadata, resolve_faults=True, sort_keys=True):
...
+            toml_doc[key] = sorted(value)
+        elif isinstance(value, dict):
+            toml_doc[key] = metadata_dict_to_toml(value, resolve_faults)
+        elif isinstance(value, Fault):
+            if resolve_faults:
+                toml_doc[key] = value.value
+            else:
+                toml_doc[key] = yellow(value._repr_first())
+        elif value is None:
+            toml_doc[key] = "!none:"
```
Kunsi: TODO discuss (bundlewrap itself does not provide this magic string by default)

Kunsi (commented):
Otherwise this is good to merge i think. Lets discuss in coworking tomorrow.

bundlewrap/metadata.py
sschiller-seibert: Todo: Set a comment instead of magic string here. Use the key somewhere inside the comment

sschiller-seibert: Comments were added in my latest commit.

phosei:
I gave this a shot, but the colors appear to be broken:

```
$ bw metadata -t myfancynode -k nginx
[metadata.nginx]
"\e[32mmodules\e[0m" = []
"\e[34mserver_names_hash_bucket_size\e[0m" = 128
```

There appear to be literal `\e` in here instead of a byte with value `0x1B`. 🤔

FTR, I merged in the current `main` – maybe this broke in the meantime?

Kunsi:
I'd propose merging this without colour support (apart from marking unresolved faults) and noting this in the help text. We can always add colours later, but i'd really like to have this and would rather have no colours than keeping this sitting around for longer.

Ideally we'd also make `metadata_to_toml()` code-compatible to `metadata_to_json()` (aka do not do `{"metadata": metadata}` in there, but instead do that in [`bw metadata`](https://github.com/bundlewrap/bundlewrap/pull/915/changes#diff-33d6f0fe61d841187bfa189c2bff29e824de8b5f4caa3a5466c9801e2c0fca78R185). That way people could also use this to write configuration files etc.

phosei:
→ https://github.com/bundlewrap/bundlewrap/pull/992

PR #920, opened 2026-01-15
`bw lock add`: add verification for --items

### LINKED ISSUE #921: bw lock does not check the items option
When using
` bw lock add -e 5d $node -i bundle:posfix`
the lock is set despite a tpyo in the -i option. This fails silently and despite the Lock is set, effectively it does not work because bw does ignore bundle:posfix. This could be avoided if bw checks the -i option for validity (i think there are more places where this could be the issue)

fkusei:
fixes #921

### DIFF
--- a/bundlewrap/cmdline/lock.py
+++ b/bundlewrap/cmdline/lock.py
@@ -7,7 +7,7 @@ from ..lock import (
     softlock_remove,
     softlocks_to_table,
 )
-from ..utils.cmdline import get_target_nodes
+from ..utils.cmdline import get_target_nodes, verify_autoskip_selectors
 from ..utils.text import (
     bold,
     error_summary,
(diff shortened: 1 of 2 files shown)


bundlewrap/cmdline/lock.py
```diff
@@ -7,7 +7,7 @@
     softlock_remove,
     softlocks_to_table,
 )
-from ..utils.cmdline import get_target_nodes
+from ..utils.cmdline import get_target_nodes, verify_autoskip_selectors
```
fkusei, line 10: I thought about maybe renaming this function, but decided against it, because this function will check if the selectors match against `Node.covered_by_autoskip_selector()` and/or `Item.covered_by_autoskip_selector()`

PR #937, opened 2026-01-21
make BUNDLE_DIR and BUNDLE_DATA_DIR available as variables in bundles

fkusei:
cc #935

### DIFF
--- a/bundlewrap/bundle.py
+++ b/bundlewrap/bundle.py
@@ -86,6 +86,8 @@ class Bundle:
             base_env = {
                 'node': self.node,
                 'repo': self.repo,
+                'BUNDLE_DIR': self.bundle_dir,
+                'BUNDLE_DATA_DIR': self.bundle_data_dir,
             }
 
             # TODO prevent access to node.metadata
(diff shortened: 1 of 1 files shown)


phosei:
Hm, I don’t have strong opinions on this. If it’s considered useful, then let’s merge it. (If this was a private project of mine, I wouldn’t merge it, because building paths using `repo.path` works just as fine.)

None of these magic variables are documented anywhere, are they?

(Do we have strong opinions on uppercase/lowercase? Maybe deprecate `repo` and `node`, and require them to be `REPO` and `NODE` in bw 6? Maybe do this for all these magic variables?)

Kunsi:
> Hm, I don’t have strong opinions on this. If it’s considered useful, then let’s merge it. (If this was a private project of mine, I wouldn’t merge it, because building paths using `repo.path` works just as fine.)

I think it would be helpful for beginners to have this - personally i don't need it either. See linked discussion.

> None of these magic variables are documented anywhere, are they?

Not that i know of.

> (Do we have strong opinions on uppercase/lowercase? Maybe deprecate `repo` and `node`, and require them to be `REPO` and `NODE` in bw 6? Maybe do this for all these magic variables?)

Sounds like a plan.

phosei:
Alright:

> > None of these magic variables are documented anywhere, are they?
> 
> Not that i know of.

https://github.com/bundlewrap/bundlewrap/issues/988

> 
> > (Do we have strong opinions on uppercase/lowercase? Maybe deprecate `repo` and `node`, and require them to be `REPO` and `NODE` in bw 6? Maybe do this for all these magic variables?)
> 
> Sounds like a plan.

https://github.com/bundlewrap/bundlewrap/issues/989

phosei (approved):

PR #957, opened 2026-02-18
deps: add option to items to skip inheriting tags

MaZderMind:
this feature was introduced in bw5 and is quite useful for most cases, but there are constructs that require the app to be up, in in order to perform a setup-task before the app is restarted later. These constructs became impossible with the new tag inheritance. This PR allows to skip the inheritance for these corner cases

### DIFF
--- a/tests/integration/bw_items.py
+++ b/tests/integration/bw_items.py
@@ -403,6 +403,7 @@ def test_bw_items_invocation_single_item_attrs(tmpdir):
 attribute\tvalue
 after\t[]
 before\t[]
+canned_actions_inherit_tags	True
 cascade_skip\tTrue
 comment\tNone
 error_on_missing_fault\tFalse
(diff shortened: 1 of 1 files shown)


bundlewrap/deps.py
```diff
@@ -193,10 +193,10 @@ def _inject_canned_actions(items):
     for item in items:
         for canned_action_name, canned_action_attrs in item.get_canned_actions().items():
             canned_action_id = f"{item.id}:{canned_action_name}"
-            canned_action_attrs.update({
-                'triggered': True,
-                'tags': item.tags,
-            })
+            canned_action_attrs.update({'triggered': True})
+            print('item', item, item.inherit_tags)
```
fkusei: oops

MaZderMind: …like it never happened. I was kind'a frustrated after 2 Days work of trying to fix he dependency loop on the other end :D

bundlewrap/items/__init__.py
```diff
@@ -44,6 +44,7 @@
     'triggers': set(),
     'unless': "",
     'when_creating': {},
+    'inherit_tags': True,
```
fkusei: hmmm, maybe `canned_actions_inherit_tags`?

Also needs docs 🥴

MaZderMind: Yea, let's see if this fixes our problems on the other end first

Issue #956, opened 2026-02-17
Option to "validate" secrets from secrets.cfg


fkusei:
When rotating a secret in `.secrets.cfg`, you face the issue of having to distribute the new secret to everyone, ideally have them all switch at the same time, so stuff is always consistent.

In our repo, we have this hook:
```
def apply_start(repo, target, nodes, interactive=False, **kwargs):
    from_bw = repo.vault.password_for('this is a secret')
    from_pw_manager = company_password_manager.get('bw secret verifier')

    if str(from_bw) != str(from_pw_manager):
        raise GracefulApplyException('secret does not match')
```

This allows admins to update the secret in the company password manager and force everyone to switch to the new secret.

However we have the issue that sometimes, users get missed when notifying about the new secret, which leads to support requests and extra hassle. This is especially annoying if the apply the user is running doesn't even require that secret.

**Bundlewrap should provide a way to "validate" a secret for correctness, but only if that secret is actually used in this apply/verify/whatever operation. This ensures users will only get prompted to update their secret if they actually need it.**

PR #962, opened 2026-02-26
Add validators

Kunsi:
This can be used to verify secret keys or the bundlewrap version actually match what you expect them to be.

Implements #956

### DIFF
--- a/bundlewrap/repo.py
+++ b/bundlewrap/repo.py
@@ -17,7 +17,7 @@ if version_info >= VERSION_NEW_PACKAGING:
 else:
     from pkg_resources import DistributionNotFound, require, VersionConflict  # needs setuptools
 
-from . import items, VERSION_STRING
+from . import items, VERSION, VERSION_STRING
 from .bundle import FILENAME_ITEMS
 from .exceptions import (
     NoSuchGroup,
@@ -52,6 +52,7 @@ FILENAME_GROUPS = "groups.py"
 FILENAME_MAGIC_STRINGS = "magic_strings.py"
 FILENAME_NODES = "nodes.py"
 FILENAME_REQUIREMENTS = "requirements.txt"
+FILENAME_VALIDATORS = "validators.py"
 
 HOOK_EVENTS = (
     'action_run_end',
@@ -280,6 +281,7 @@ class Repository(MetadataGenerator):
         self.node_dict = {}
         self.node_attribute_functions = {}
         self.magic_string_functions = {}
+        self.validators = {}
         self._get_all_attr_code_cache = {}
         self._get_all_attr_result_cache = {}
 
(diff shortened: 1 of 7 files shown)


tests/integration/validators.py
```diff
@@ -0,0 +1,75 @@
...
+        nodes = {
+            'node1': {
+                'metadata': {
+                    'foo': vault.password_for('node1'),
+                },
+            },
+        }
+        """)
+        )
+    with open(join(tmpdir, "validators.py"), "w") as f:
```
Kunsi, line 27: Running without `validators.py` is impicitly tested in the other integration tests.

bundlewrap/secrets.py
```diff
@@ -70,6 +71,30 @@ def random_key():
...
+            if key not in self.validator_locks:
+                self.validator_locks[key] = Lock()
+
+        # We need this lock to ensure we only run validators once per
+        # key. After all, validators might take a long time or use a
+        # lot of resources depending on what users are doing.
+        with self.validator_locks[key]:
+            if key in self.validator_cache:
+                if not self.validator_cache[key]:
+                    raise FaultUnavailable(_("validator for key '{}' failed").format(key))
```
Kunsi, line 89: To be discussed: is `FaultUnavailable` maybe too soft? Might get missed in big `bw apply` runs?

bundlewrap/repo.py
```diff
@@ -498,6 +500,36 @@ def magic_string(func):
...
+        Runs a validator. Calling code MUST catch raised ValidatorError
+        exceptions and properly handle them. Calling code also SHOULD
+        cache the result if it's expected that validators run more than
+        once during a `bw` run.
+        """
+        try:
+            func = self.validators[validator_name]
+        except KeyError:
+            # if the function is not defined, return early
+            return
```
Kunsi, line 528: Discuss: Make this mandatory? (amd provide defaults for of FILENAME_VALIDATORS doesn't exist)

docs/content/repo/validators.py.md
```diff
@@ -0,0 +1,33 @@
+# Validators
```
phosei, line 1: Took me a good second to wrap my head around how to use validators. Maybe a real-life example would help. 🤔

bundlewrap/secrets.py
phosei: Yeah, I think `FaultUnavailable` is probably too soft.

-   Key not present (like because you’re a dev and not an admin)? → `FaultUnavailable`
-   Key present but _wrong_? This should be an error.

bundlewrap/repo.py
phosei: Make what exactly mandatory? 😅

trehn: not sure if silently ignoring typos will fit most users' expecations

PR #980, opened 2026-04-20
items/svc_systemd: Mask unit only after disabling it

xje4:
When a systemd unit is to be disabled and masked, that needs to happen in this order. If a unit has been masked, systemd refuses to disable it and the symlink in the .wants-directory stays. Same thing applies in the reverse order when unmasking and enabling a unit again. It first has to be unmasked before it can be enabled.

With the current order in which bundlewrap processes a service (first mask/unmask, then enable/disable) this creates a state on the remote system in which the unit is masked but still "enabled". For the unit itself, this has no impact, since a masked unit instantly fails when it's tried to be started (although it's still a bit ugly to have those "orphaned" symlinks in the .wants-directory laying around).
However, there's a bigger impact when the unit has defined `Also` units like NetworkManager. When NetworkManager is disabled regularly, it pulls in `NetworkManager-wait-online` which is then also disabled. When however NetworkManager is masked (and then cannot be disabled anymore), `NetworkManager-wait-online` stays enabled and (since _that_ unit hasn't been masked) is started normally which causes it to wait for NetworkManager which is never gonna be active.

Therefore, fix the order in which systemd-units are masked, unmasked, enabled and disabled so the masking is done at the end while the unmasking stays at the top. This unfortunately requires to do the `if 'masked' in status.keys_to_fix` twice, once for unmasking at the start and again for the masking at the end.

### DIFF
--- a/bundlewrap/items/svc_systemd.py
+++ b/bundlewrap/items/svc_systemd.py
@@ -87,11 +87,8 @@ class SvcSystemd(Item):
         return state
 
     def fix(self, status):
-        if 'masked' in status.keys_to_fix:
-            if self.attributes['masked']:
-                svc_mask(self.node, self.name)
-            else:
-                svc_unmask(self.node, self.name)
+        if 'masked' in status.keys_to_fix and not self.attributes['masked']:
+            svc_unmask(self.node, self.name)
 
         if 'enabled' in status.keys_to_fix:
             if self.attributes['enabled']:
@@ -104,6 +101,9 @@ class SvcSystemd(Item):
                 svc_start(self.node, self.name)
             else:
                 svc_stop(self.node, self.name)
+
+        if 'masked' in status.keys_to_fix and self.attributes['masked']:
+            svc_mask(self.node, self.name)
 
     def get_canned_actions(self):
         return {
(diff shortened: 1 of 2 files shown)


xje4:
Looks like DeepSource isn't happy with my double-`if 'masked'` condition, yet I believe for a method of this size and structure, that's the least ugly solution.
Anyway, what alternative code structure would you choose for this change?

bundlewrap/items/svc_systemd.py
```diff
@@ -88,9 +88,7 @@ def expected_state(self):
 
     def fix(self, status):
         if 'masked' in status.keys_to_fix:
-            if self.attributes['masked']:
-                svc_mask(self.node, self.name)
-            else:
+            if not self.attributes['masked']:
```
fkusei: You could combine those `if`s to one, that way deepsource will probably be happier:
```python
if 'masked' in status.keys_to_fix and not self.attributes['masked']:
```
(same below just with the inverted `and`)

xje4: Oh, yeah. I wouldn't have thought that such a check could be circumvented so easily. Thanks

Kunsi (approved):
LGTM, will merge after 5.0.4 is out, 5.1.0 should be soon as well.

phosei:
@xje4 Looks good to me as well, thanks!

Could you please sign the CAA? (Step 7: https://docs.bundlewrap.org/misc/contributing/#contributing-code) It’s probably better this way, you’ve contributed a number of things now. ✌️

xje4:
Done

phosei:
Great, thank you!

PR #992, opened 2026-05-26
add toml output to bw metadata

phosei:
This is @sschiller-seibert’s PR https://github.com/bundlewrap/bundlewrap/pull/915 with two extra commits (I can’t push on her branch).

First, `metadata_to_toml()` now works like `metadata_to_json()`: No extra `metadata` gets added. This addresses the second half of @fkusei’s comment here https://github.com/bundlewrap/bundlewrap/pull/915#issuecomment-4485839606.

Then, colors have been fixed. Apparently, tomlkit returns the literal string `\e` for escape bytes?

image

Works for me now:

image

But please run additional tests on your end. Do colors really work for you now? If they don’t, we should drop the color feature from TOML output for now (as @fkusei suggested), so we can get this merged soon.

### DIFF
--- a/bundlewrap/metadata.py
+++ b/bundlewrap/metadata.py
@@ -1,12 +1,16 @@
 from copy import copy
 from hashlib import sha256
-from json import dumps, JSONEncoder
+from json import JSONEncoder, dumps
+from types import NoneType
+
+from tomlkit import document as toml_document
+from tomlkit import dumps as tomlkit_dumps
 
 from .exceptions import RepositoryError
 from .utils import Fault
 from .utils.dicts import ATOMIC_TYPES, map_dict_keys, merge_dict, value_at_key_path
-from .utils.text import force_text, mark_for_translation as _, yellow
-
+from .utils.text import force_text, yellow
+from .utils.text import mark_for_translation as _
 
 METADATA_TYPES = (  # only meant for natively atomic types
     bool,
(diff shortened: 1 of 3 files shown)


fkusei:
Can confirm TOML with colours works here (on MacOS).

I'm still not sure if we should merge this as it is right now, because we get a lot of unneeded `"` here (example):
```toml
[metadata."docker-engine"]
"manage_var_lib_containerd" = true
```

bundlewrap/metadata.py
```diff
@@ -1,12 +1,16 @@
 from copy import copy
 from hashlib import sha256
-from json import dumps, JSONEncoder
+from json import JSONEncoder, dumps
+
+from tomlkit import comment
```
fkusei, line 5: unused import

bundlewrap/metadata.py
```diff
@@ -336,6 +340,31 @@ def metadata_to_json(metadata, resolve_faults=True, sort_keys=True):
...
+                toml_doc.add(key, yellow(value._repr_first()))
+        elif value is None:
+            toml_doc.add(key, "None")
+            toml_doc[key].comment("# was None, but toml can't natively handle NoneType objects")
+        else:
+            toml_doc.add(key, value)
+    return toml_doc
+
+
+def metadata_to_toml(metadata, resolve_faults=True, sort_keys=True):
```
fkusei, line 365: `sort_keys` is not used here.

Issue #713, opened 2022-06-20
[routeros] KeyErrors when generating sdict


fkusei:
If there's a port which is currently disabled, but `cdict` is as follows:
```json
{
  "disabled": "false",
  "hw": "true"
}
```
Bundlewrap will generate a `KeyError('hw')` (see blelow), because that key isn't present in the `sdict` for the disabled port.

I'm not sure wether we should just use `.get()` in `keys_to_fix()` or handle those edge cases differently.

Full error message:
```
✘ wi-5.hw.sw-mn  routeros  routeros:/interface/bridge/port?interface=ether23  failed 
│
├─ ('/interface/bridge/port/print', '?=interface=ether23', '?#&') (return code: None)
│ 
│ ({'.id': '*16', '.nextid': '*17', 'interface': 'ether23', 'bridge': 'bridge', 'priority': '0x80', 'path-cost': 10, 'internal-path-cost': 10, 'edge': 'auto', 'point-to-point': 'auto', 'learn': 'auto', 'horizon': 'none', 'auto-isolate': False, 'restricted-role': False, 'restricted-tcn': False, 'pvid': 1, 'frame-types': 'admit-all', 'ingress-filtering': False, 'unknown-unicast-flood': True, 'unknown-multicast-flood': True, 'broadcast-flood': True, 'tag-stacking': False, 'bpdu-guard': False, 'trusted': False, 'multicast-router': 'temporary-query', 'fast-leave': False, 'status': 'inactive', 'inactive': True, 'dynamic': False, 'disabled': True},)
╵
│   File "/home/fkunsmann/.venv/isac/lib/python3.10/site-packages/bundlewrap/concurrency.py", line 133, in run
│     result = self._get_result()
│   File "/home/fkunsmann/.venv/isac/lib/python3.10/site-packages/bundlewrap/concurrency.py", line 73, in _get_result
│     raise exception
│   File "/usr/lib/python3.10/concurrent/futures/thread.py", line 58, in run
│     result = self.fn(*self.args, **self.kwargs)
│   File "/home/fkunsmann/.venv/isac/lib/python3.10/site-packages/bundlewrap/items/__init__.py", line 554, in apply
│     status_before = self.cached_status
│   File "/home/fkunsmann/.venv/isac/lib/python3.10/site-packages/bundlewrap/utils/__init__.py", line 36, in cache_wrapper
│     return_value = prop(self)
[...]

PR #999, opened 2026-06-24
fix: Add empty string as default for routeros `actual_state`

CommanderRedYT:
The RouterOS API does not return attributes that are not set sometimes. By defaulting all expected keys to empty-string, we can mitigate this problem.

The problem actually appears when applying, as some code expects all keys from `expected_state` to also exist in `actual_state`, which was not given before.

### DIFF
--- a/bundlewrap/items/routeros.py
+++ b/bundlewrap/items/routeros.py
@@ -97,6 +97,12 @@ class RouterOS(Item):
         if state:
             # API doesn't return comment at all if emtpy
             state.setdefault('comment', '')
+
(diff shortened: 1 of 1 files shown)


Kunsi:
FTR, something like that was already discussed here: https://github.com/bundlewrap/bundlewrap/issues/713

(i'm not saying we shouldn't do this though)

CommanderRedYT:
I'd say if its done for the `comment` key then why not for the others. I also tried adding if-statements with `hasattr(state, key)` everywhere before I came up with this, but that turned out to make more bugs.

PR #1001, opened 2026-07-13
`bw lock add`: Allow users to set a default for `--skip-item-verification`

phosei:
This feature got introduced in bw 5.0.0. Problem is, it can be very costly because it requires bw to generate all items for a node. Nobody cares on small nodes, but on large nodes with lots of metadata reactors, this can take several *minutes*.

So, allow users to customize this by setting `$BW_LOCK_ADD_SKIP_ITEM_VERIFICATION`.

Here's a demo (with a little patch that only prints the value of `args['skip_item_verification']` without actually locking anything):

    $ bw lock add bignode -e 1m
    False
    $ bw lock add bignode -e 1m --skip-item-verification
    True
    $ bw lock add bignode -e 1m --no-skip-item-verification
    False

    $ BW_LOCK_ADD_SKIP_ITEM_VERIFICATION=0 bw lock add bignode -e 1m
    False
    $ BW_LOCK_ADD_SKIP_ITEM_VERIFICATION=0 bw lock add bignode -e 1m --skip-item-verification
    True
    $ BW_LOCK_ADD_SKIP_ITEM_VERIFICATION=0 bw lock add bignode -e 1m --no-skip-item-verification
    False

    $ BW_LOCK_ADD_SKIP_ITEM_VERIFICATION=1 bw lock add bignode -e 1m
    True
    $ BW_LOCK_ADD_SKIP_ITEM_VERIFICATION=1 bw lock add bignode -e 1m --skip-item-verification
    True
    $ BW_LOCK_ADD_SKIP_ITEM_VERIFICATION=1 bw lock add bignode -e 1m --no-skip-item-verification
    False

Requires Python 3.9+ because of `BooleanOptionalAction`.

### DIFF
--- a/bundlewrap/cmdline/parser.py
+++ b/bundlewrap/cmdline/parser.py
@@ -1,4 +1,5 @@
-from argparse import ArgumentParser, RawTextHelpFormatter, SUPPRESS
+from argparse import (ArgumentParser, BooleanOptionalAction,
+                      RawTextHelpFormatter, SUPPRESS)
 from os import environ, getcwd
 from os.path import join
 
@@ -581,7 +582,8 @@ bundle:my_bundle  # items in this bundle
     )
     parser_lock_add.add_argument(
         "--skip-item-verification",
-        action='store_true',
+        action=BooleanOptionalAction,
+        default=bool(int(environ.get('BW_LOCK_ADD_SKIP_ITEM_VERIFICATION', 0))),
         dest='skip_item_verification',
         help=_("disable verification of item selector (-i)"),
     )


bundlewrap/cmdline/parser.py
```diff
@@ -581,7 +582,8 @@ def build_parser_bw():
     )
     parser_lock_add.add_argument(
         "--skip-item-verification",
-        action='store_true',
+        action=BooleanOptionalAction,
+        default=bool(environ.get('BW_LOCK_ADD_SKIP_ITEM_VERIFICATION')),
```
phosei: Should we add another `int()` here? Allow users to set this to `1` or `0`? At the moment, it’s only “set” or “unset”.

phosei: Yeah, that’s probably better.

```
$ bw lock add bignode -e 1m
False
$ bw lock add bignode -e 1m --skip-item-verification
True
$ bw lock add bignode -e 1m --no-skip-item-verification
False
```

```
$ BW_LOCK_ADD_SKIP_ITEM_VERIFICATION=1 bw lock add bignode -e 1m
True
$ BW_LOCK_ADD_SKIP_ITEM_VERIFICATION=1 bw lock add bignode -e 1m --skip-item-verification
True
$ BW_LOCK_ADD_SKIP_ITEM_VERIFICATION=1 bw lock add bignode -e 1m --no-skip-item-verification
False
```

```
$ BW_LOCK_ADD_SKIP_ITEM_VERIFICATION=0 bw lock add bignode -e 1m
False
$ BW_LOCK_ADD_SKIP_ITEM_VERIFICATION=0 bw lock add bignode -e 1m --skip-item-verification
True
$ BW_LOCK_ADD_SKIP_ITEM_VERIFICATION=0 bw lock add bignode -e 1m --no-skip-item-verification
False
```

Kunsi:
I'm about 53% certain we should just remove item verification for `bw lock add`, tbh.

I originally only wanted this for `bw apply`, then someone (can't remember who) wanted to also have that for `bw lock add`, but i'm more and more certain adding this there was the wrong move.

Kunsi:
Another thought: Should this maybe be a list of nodes where item verification is skipped by default?

(This is how we end up with a `bundlewrap.cfg` in the repo, isn't it?)

phosei:
> I'm about 53% certain we should just remove item verification for `bw lock add`, tbh.
> 
> I originally only wanted this for `bw apply`, then someone (can't remember who) wanted to also have that for `bw lock add`, but i'm more and more certain adding this there was the wrong move.

I dug it up (corporate chat) but obviously can’t publish that conversation here. (Note to self: `~/gfx/screenshots/2026-07-16--07-42-48-450574.png` and `~/gfx/screenshots/2026-07-16--07-42-48-450574.png.README`.)

It was requested by a user as a safe-guard against typos. It’s *probably* not a bad idea to have this … ?

> Another thought: Should this maybe be a list of nodes where item verification is skipped by default?
> 
> (This is how we end up with a `bundlewrap.cfg` in the repo, isn't it?)

🙈🙈🙈

Kunsi (approved):

PR #1002, opened 2026-07-15
items/directories: add pre_purge_command

Kunsi:
Allows users to do stuff like "Disable systemd units prior to purging unused unit files" etc.

### DIFF
--- a/bundlewrap/items/directories.py
+++ b/bundlewrap/items/directories.py
@@ -49,6 +49,7 @@ class Directory(Item):
         'mode': "0755",
         'owner': "root",
         'purge': False,
+        'pre_purge_command': None,
     }
     ITEM_TYPE_NAME = "directory"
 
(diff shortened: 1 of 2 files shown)


bundlewrap/items/directories.py
```diff
@@ -98,6 +99,11 @@ def fix(self, status):
             return
 
         for path in status.actual_state.get('paths_to_purge', set()):
+            if self.attributes['pre_purge_command']:
+                self.run(self.attributes['pre_purge_command'].format(
+                    path=quote(path),
+                    file=quote(path.split('/')[-1]),
```
phosei, line 105: I wonder if we really need this. Can’t the `pre_purge_command`s do that themselves?

Kunsi: In theory, yes, but it felt better to include this here because it will significantly improve readability and UX if we have it.

phosei (approved):

PR #1006, opened 2026-09-04
bw verify: skip items with unavailable Faults instead of tracebacks

CroneKorkN:
`bw apply` already skips an item when a Fault becomes unavailable while computing its expected state (e.g. inside a template). `bw verify` only handled Faults detected at item creation; a Fault surfacing later ended up in the generic exception handler and printed a full traceback as "unable to get status".

Treat it like the early check: print a yellow "Fault unavailable" line and count the item as unknown.

Claude-Session: https://claude.ai/code/session_01KyF5kCuP8utyDgND7enAB8

### DIFF
--- a/bundlewrap/cmdline/diff.py
+++ b/bundlewrap/cmdline/diff.py
@@ -2,11 +2,11 @@ from copy import copy
 from difflib import unified_diff
 from sys import exit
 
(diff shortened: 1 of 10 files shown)


Kunsi:
This was implemented without LLM-generated code in #1007

Issue #1008, opened 2026-09-09
Per node/per group secrets for generate and encrypt


sschiller-seibert:
Currently, all password derivation functions like `repo.vault.password_for()` use the repo wide key 'generate'. This becomes a problem once the userbase of a repo gets larger and not every user is allowed to carry the big keys to the kingdom. 

Metadata reactors that read the name of a generate key from metadata are possible, but every every bundle dev has to know about it and consistently use it correctly. 

```
def django_secret(metadata):
    return {
        'importantapp': {
            'secrets': {
                'django': {
                    'secret': repo.vault.password_for('django secret on {}'.format(node.name), key=metadata.get('generate_key', 'generate'))

                },
            },
        },
    }

```

Having per node or per group secrets outside of metadata that automatically get chosen for all functions in `repo.vault` would make this much easier. Maybe even completely replace the repo-wide `repo.vault` with node specific `node.vault`.

Additinally, it should be possible to use `bw pw -e` in the context of a node like `bw debug` already is.

All of this will likely be a breaking change.

PR #1009, opened 2026-09-09
node.vault: per-node and per-group secret keys

CroneKorkN:
Refs #1008

### Motivation

`repo.vault.password_for()` and friends always use the repo-wide `generate` key (and `encrypt` for
encryption). Once several teams share a repository and not everyone should hold every key, every
bundle author has to remember to thread a key name through metadata into every vault call. This PR
makes the key a property of the node instead.

### What's new

* **Node/group attributes `generate_key` and `encrypt_key`** (`str`, default `None`). They inherit
  like every other group attribute (node > subgroup > supergroup > default) and work in TOML nodes
  and groups at top level.
* **`node.vault`** — same API as `repo.vault` (`password_for`, `human_password_for`,
  `random_bytes_as_base64_for`, `encrypt`, `encrypt_file`, `decrypt`, `decrypt_file`,
  `decrypt_file_as_base64`, `cmd`). Without an explicit `key=`, deriving uses `node.generate_key`
  and encrypting uses `node.encrypt_key`. Decryption is unchanged: the `keyname$` prefix decides,
  and ciphertext without prefix is still decrypted with `encrypt` (such ciphertext predates key
  prefixes and was necessarily produced with that key). `key=` always wins. Nodes without the
  attributes produce the same values and use the same keys as `repo.vault` (their `Fault` objects
  additionally carry the node name in their identity, so they are not `==` to `repo.vault` Faults).
* **`bw pw -n NODE`** — derive and encrypt with the node's keys; `-k` still overrides, decryption is
  unchanged.
* **Magic strings receive the node**: a function registered in `magic_strings.py` gets `node=` if
  its signature has a `node` parameter (or `**kwargs`), so `"!password_for:foo"` in a TOML node can
  use `node.vault`. Group magic strings never get a node. Also: the argument part may be empty
  (`"!none:"`) and `@magic_string(name="...")` registers names that aren't valid identifiers.
* **`secret_key_use` hook** gets `node=` when the key is used via `node.vault` (omitted for
  `repo.vault`, so hooks following the documented `**kwargs` convention keep working).

The first commit is independent of the feature: it completes the `bw pw` section of the CLI docs
(`-d` and `-b` examples, `-k`/`-l`/`-f`), which the new `-n` paragraph builds on.

### What this deliberately does not do

#1008 asks for keys that are "automatically chosen for all functions in `repo.vault`". This PR does
not change `repo.vault`: existing calls keep using the repo-wide keys, and per-node keys are opt-in
via `node.vault`. That keeps the change additive and lets a repository migrate call by call. A
follow-up could expose `node.vault` under a shorter name in bundle environments if that turns out
to be the missing convenience.

### Compatibility

Purely additive. `repo.vault`, its defaults and its Fault identities are unchanged; existing
`magic_strings.py` functions and `secret_key_use` hooks keep working; `bw pw` without `-n` behaves
exactly as before. One edge case to be aware of: because the magic string argument may now be
empty, `"!unknown:"` (unregistered name, empty argument) now raises `InvalidMagicStringException`
instead of being passed through as a literal string.

### Tests

Unit tests for attribute inheritance and the node-aware magic string conversion; integration tests
(no SSH needed) for `node.vault` key resolution, encryption/decryption with and without prefix,
`bw pw -n`, magic strings with `node`, and the hook. Full suite run locally: all SSH-free tests pass,
the remaining failures are the SSH-dependent apply/run/verify tests that only run in CI.

We are migrating our internal repository (~1300 nodes, ~90 bundles touched, ~160 vault calls) in
parallel and will report back on the issue.

### Open questions for maintainers

* Release: this is 5.x-compatible; happy to move to the 6.0 milestone if you'd rather ship it there
  (#1008 is labelled `next-major-release`).
* Naming: `generate_key`/`encrypt_key` mirror the `.secrets.cfg` section names; alternatives welcome.
* Changelog: I added an "Unreleased" heading; happy to drop it and let the entry be folded into the
  release commit instead, whichever you prefer.
* `secret_key_use` concurrency (pre-existing, not changed here): the re-entrancy guard in
  `SecretProxy.__hook` is per key and shared between threads, so when two workers resolve Faults with
  the same key at the same time, only one of them reaches the hook. The `node=` argument makes this
  more visible, so the docs now say the hook is not a reliable per-use guard. A thread-local guard
  keyed by `(key, node)` would fix it, but it changes how often the hook fires for `repo.vault` too,
  so I left that decision to you; happy to add it in this PR or a follow-up.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)


bundlewrap/utils/magic_strings.py
```diff
@@ -27,15 +57,18 @@ def convert_magic_strings(repo, obj):
                 raise InvalidMagicStringException(func_name)
             else:
                 with error_context(magic_string=func_name):
-                    obj = func(func_args)
+                    if node is not None and _accepts_node(func):
```
trehn, line 60: introspection here feels a bit much, but catching `TypeError` instead is too broad, so I don't really have a better idea

docs/content/repo/groups.py.md
```diff
@@ -38,7 +38,7 @@ This section is a reference for all possible attributes you can define for a gro
 	    },
 	}
 
-Note that many attributes from [nodes.py](nodes.py.md) (e.g. `bundles`) may also be set at group level, but aren't explicitly documented here again.
+Note that many attributes from [nodes.py](nodes.py.md) (e.g. `bundles`, `generate_key`, `encrypt_key`) may also be set at group level, but aren't explicitly documented here again.
```
trehn, line 41: no need to name them here

bundlewrap/utils/magic_strings.py
```diff
@@ -1,14 +1,44 @@
...
+MAGIC_STRINGS_PATTERN = compile(r"^!(" + MAGIC_STRING_NAME + r"):(.*)$")
 
+_ACCEPTS_NODE = {}
 
-def convert_magic_strings(repo, obj):
+
+def _accepts_node(func):
+    """
+    True if func can be called with node=..., i.e. it has a `node`
+    parameter or takes **kwargs. Cached, signature() is not cheap.
```
fkusei, line 19: Should we maybe wrap this in a `Lock()` so we can ensure to only run this once per run?

bundlewrap/group.py
```diff
@@ -95,6 +100,13 @@
 }
 
 
+def validate_secret_key_attrs(attributes):
+    # an empty string would silently select the default key
```
fkusei, line 104: Not sure if i like this. TOML does not have a `None` option, so we should have some way to override the key to "just use the default" in TOML.

bundlewrap/node.py
```diff
@@ -562,6 +569,12 @@ def __init__(self, name, attributes=None, repo=None):
     def __getattr__(self, name):
         with suppress(KeyError):
             return self._dynamic_attribute_cache[name]
+        if name.startswith("_") and name[1:] in GROUP_ATTR_DEFAULTS:
+            # the backing fields are only set at the end of __init__
+            raise RepositoryError(_(
+                "node attribute '{attr}' of node '{node}' is not available yet "
+                "while the node is being loaded (use node.vault instead)"
```
fkusei, line 576: This catches all kinds of errors, but only mentions `node.vault` in the error message which might be confusing for users.

bundlewrap/secrets.py
```diff
@@ -407,3 +417,91 @@ def random_bytes_as_base64_for(self, identifier, key='generate', length=32):
             key=key,
             length=length,
         )
+
+
+class NodeVault:
```
fkusei, line 422: To be discussed: If we have node-specific secrets now, should this also add the node name to the generator as a seed, so `NodeA.vault.password_for('foo')` and `NodeB.vault.password_for('foo')` generate different passwords?

docs/content/guide/secrets.md
```diff
@@ -129,6 +129,26 @@ The encrypted data will be prefixed by `yourkeyname$...` to indicate that the ke
 
 <br>
 
+### node.vault: per-node and per-group keys
+
+Instead of passing `key=` everywhere, name the keys a node should use with the [`generate_key` and `encrypt_key`](../repo/nodes.py.md#generate_key) attributes on the node or one of its groups:
+
+	# groups/team_a.toml
```
fkusei, line 136: Unclear, this lists a group but the code block below is using `node`, better make this a node here as well.

docs/content/guide/toml.md
```diff
@@ -109,8 +109,38 @@ mysecret = "!decrypt:encrypt$gAAAAABo90x3H..."
 ```
 
 The part between `!` and `:` is used as the function name, everything after the `:` will be passed
-as argument to the called function. Bundlewrap will raise `InvalidMagicStringException` if the
-function cannot be found.
+as argument to the called function (this may be an empty string, e.g. `"!none:"`). Bundlewrap will
```
fkusei, line 112: > (this may be an empty string, for example if you create a function to produce a literal `None`)

We should avoid listing a magic string if bw doesn't ship with it.

docs/content/guide/toml.md
```diff
@@ -109,8 +109,38 @@ mysecret = "!decrypt:encrypt$gAAAAABo90x3H..."
...
+[`node.vault`](secrets.md#nodevault-per-node-and-per-group-keys) to use the node's own keys:
+
+```python
+@magic_string
+def password_for(identifier, node=None):
+    # `node` is None for magic strings in groups
+    return (node.vault if node else vault).password_for(identifier)
+```
+
+The node is still being loaded at this point. `node.name` and the lazy `node.vault` functions
```
fkusei, line 137: This paragraph should be removed because it's just needed for some AI to understand it, not for humans.

docs/content/repo/hooks.md
```diff
@@ -291,13 +291,14 @@ Called each time a `bw run` command finishes.
 
 **`secret_key_use(repo, key, **kwargs)`**
 
-Called once for every usage of a key from [`.secrets.cfg`](../guide/secrets.md).
-Basically this means this hook will get called every time a Fault generated by `repo.vault` is resolved.
+Called when a key from [`.secrets.cfg`](../guide/secrets.md) is used, i.e. when a Fault generated by `repo.vault` or `node.vault` is resolved. It runs at most once at a time per key: other threads resolving a Fault with the same key while it is running skip it, so it cannot serve as a reliable guard for every single use of a key.
```
fkusei, line 294: This is wrong, other threads will wait until the hook is done. The only time the hook is skipped is if the hook itself calls `vault` to avoid a recursion loop.

docs/content/repo/hooks.md
```diff
@@ -291,13 +291,14 @@ Called each time a `bw run` command finishes.
...
 
 `repo` The current repository (instance of `bundlewrap.repo.Repository`).
 
 `key` The secret key that's being used
 
+`node` The node whose `node.vault` is being used (instance of `bundlewrap.node.Node`). Only present for `node.vault`, so use `kwargs.get('node')`.
```
fkusei, line 300: The default should be `None`, so `node` is always present.

docs/content/repo/hooks.md
```diff
@@ -291,13 +291,14 @@ Called each time a `bw run` command finishes.
 
 **`secret_key_use(repo, key, **kwargs)`**
```
fkusei, line 292: `node` missing here

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+# Unreleased
```
fkusei, line 1: CHANGELOG is only touched on release by a maintainer. Please do not fill it before a release is due.

fkusei (changes requested):
In general:

- Add comments to remove the "node is an optional argument to a magic string" in 6.0.0
- In documentation try to limit the line length to 80 characters, maximum is 100
- Your commit messages contain AI references. We do not want "i just let my AI generated code without checking it" commits in this repository. If you checked the code and are confident it does what it should do, then you are the author, not some AI, so remove the AI references.

